PR #137, opened 2025-07-11
Update API key precedence and remove API key storing in global env

carlosejimenez:
Note: Users should remove `api_key` keys from the global config to avoid confusion.


klieret:
Awesome, let me fix the tests and let's gooooo

PR #170, opened 2025-07-15
Improve clarity of user prompts in the `micro` cli

carlosejimenez:
In the confirmation prompt, "enter comment/command" is slightly ambiguous. This PR proposes a reword to the interactive user prompts to make available commands clearer.

### Changes
- Updated confirmation prompt in `ask_confirmation()` method
- Updated interruption prompt in `step()` method's KeyboardInterrupt handler

### Before/After
**Before:**
```
Execute? Enter to confirm, /h for help, or enter comment/command
> 
```

**After:**
```
Execute? Enter to confirm, or type a comment/command (/h for available commands)
> 
```


klieret:
nice, thanks!

PR #172, opened 2025-07-16
Change: Source bashrc with docker

klieret:


klieret:
I wonder if it behaves similar to `-i` though. With `-i` things work, but you get some annoying job control error messages (because `docker exec` doesn't support it). I can test tomorrow though

closji:
Hmm, yeah nvm, I think your first solution might actually be more robust.
Do we want to make the rcfile configurable at all?

klieret:
Actually the `-l` option (=`--login`) seems to work and not throw any errors, so let's go with that. It seems that typically `.profile` loads the bashrc.

PR #178, opened 2025-07-16
Feat: Provide set-key functionality

### LINKED ISSUE #174: no api key error message should explain how to set the api key
> AuthenticationError: litellm.AuthenticationError: Missing Anthropic API Key - A call is 
being made to anthropic but no key is set either in the environment variables or via 
params. Please set `ANTHROPIC_API_KEY` in your environment vars

this should also give the command for setting an environment var so that the user doesn't need to look it up

klieret:
Closes #174

### DIFF
--- a/src/microsweagent/models/litellm_model.py
+++ b/src/microsweagent/models/litellm_model.py
@@ -35,16 +35,14 @@ class LitellmModel:
             )
         ),
     )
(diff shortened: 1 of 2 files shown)


klieret:
Thanks Ofir, I'll take care of the unit test in a sec

PR #223, opened 2025-07-20
add responsive task prompt for micro command

carlosejimenez:
This updates the main `micro` task prompt with a responsive bottom toolbar.
When Escape is pressed and the prompt is ready to be submitted, we update the bottom toolbar from "Confirm with Esc, then Enter" to "Press Enter to submit". Behavior before had "Confirm with Esc, then Enter" before and after pressing Escape, leaving some ambiguity to how to submit.

### DIFF
--- a/src/microsweagent/run/micro.py
+++ b/src/microsweagent/run/micro.py
@@ -89,13 +89,11 @@ def main(
     config = yaml.safe_load(get_config_path(config_spec).read_text())
 
     if not task:
-        console.print(
-            "[bold yellow]What do you want to do?\n"
-            "[bold green]Up[/bold green]/[bold green]Down[/bold green] to bring up previous tasks or [bold green]Ctrl+R[/bold green] to search history\n"
-        )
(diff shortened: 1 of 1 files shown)


klieret:
Hmm, I feel a little bit conflicted because it adds so many lines... But if you feel like this is really confusing to people, then it migth well be worth it. Though maybe we can pull it out into a `multiline_input` utility to not add so much immediate complexity to the file?

klieret:
How about writing "Esc+Enter" (you can also press them together)? I feel like that always denotes a combination.

Can also show it in brighter color:

image

closji:
Okay! I think that's an acceptable option.
I updated the PR to be similar to your suggestion.

closji:
Screenshot 2025-07-21 at 7 36 51 PM

Okay, I made some other updates, similar to your suggestion.
What do you think about this?

klieret:
Awesome, thanks!

PR #276, opened 2025-07-24
Fix: Replace union syntax with Union import for Python compatibility

WuJunde:
Replace Python 3.10+ union syntax (|) with typing.Union for better compatibility with older Python versions.

### DIFF
--- a/src/minisweagent/run/mini.py
+++ b/src/minisweagent/run/mini.py
@@ -14,7 +14,7 @@ More information about the usage: [bold green]https://mini-swe-agent.com/latest/
 
 import os
 from pathlib import Path
-from typing import Any
+from typing import Any, Union
 
 import typer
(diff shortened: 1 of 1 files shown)


klieret:
Hi @WuJunde , really appreciate the PR ❤️ 

However, 3.9 is 3 months away from its [end of life](https://endoflife.date/python), so sticking to >= 3.10 makes sense to me, especially for a project that wants to show beautiful code. 

Sorry about that!

WuJunde:
got it, seems I'm gonna to upgrade my python🤣

klieret:
If it consoles you, you'll also get more readable exceptions by default and faster execution :)   (as a built-in python feature across all projects) 

But now that you bring it up, let me check if we're actually compatible with 3.10, I think that should work #277

PR #288, opened 2025-07-25
Fix links on the landing page

callmephilip:
A few links are broken on the landing page (due to `/latest` prefix). This PR fixes this by using relative URLs instead


callmephilip:
> Ohhh, why didn't the link checker catch that! Really appreciate you taking the time to directly fix this ❤️❤️

🫡

PR #297, opened 2025-07-25
(Minor) Show SWE-agent paper citation

john-b-yang:
With the drop down, I think the citation might be a bit easy to miss. Proposing we remove the `<details>` stuff so it's apparent :)


klieret (approved):
sgtm!

PR #298, opened 2025-07-25
Fix #144: Enable different kinds of parsers for actions

### LINKED ISSUE #144: Allow to change action parsing to XML
Maybe:

1. Make the parsing regex configurable (just move it into the config class)
2. Add some documentation

john-b-yang:
Enables specifying different kinds of action parsers via the config.
* Adds an `ActionParserMap` + `action_parser` property to the `AgentConfig` class
* Add `action_parser` key to existing `src/minisweagent/*.yaml` configs.

### DIFF
--- a/src/minisweagent/agents/default.py
+++ b/src/minisweagent/agents/default.py
@@ -11,11 +11,6 @@ from jinja2 import Template
 
 from minisweagent import Environment, Model
 
-ActionParserMap = {
-    "backticks": r"```bash\n(.*?)\n```",
-    "xml": r"<command>(.*?)</command>",
-}
-
 
 @dataclass
 class AgentConfig:
@@ -32,7 +27,7 @@ class AgentConfig:
     )
     format_error_template: str = "Please always provide EXACTLY ONE action in triple backticks."
     action_observation_template: str = "Observation: {{output}}"
-    action_parser: str = "backticks"  # Options: "backticks", "xml"
+    action_parser: str = "```bash\n(.*?)\n```"
     step_limit: int = 0
     cost_limit: float = 3.0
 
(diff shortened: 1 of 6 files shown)


john-b-yang:
I just started with the simplest fix, but very open to moving things around / making things more well documented @klieret.

I do kind of wonder, it's quite this repo will get more and more configuration requests in the future. In which cas, stuffing all the configs inside `default.py` might not be the best way to go about this?

klieret:
Thanks for opening this! 

> I do kind of wonder, it's quite this repo will get more and more configuration requests in the future. In which cas, stuffing all the configs inside default.py might not be the best way to go about this?

My thoughts exactly, let's really keep things minimal this time. 

Honestly, I'm thinking to maybe just not do this right now when we don't really have a use case for it ourselves anyway? What do you think?

I think moving forward we have a couple of options, but we can also just always add small things to the [cookbook](https://mini-swe-agent.com/latest/advanced/cookbook/)? 

Because also here, all that someone needs to do is to write a trivial subclass that just changes this one method...

I think going forward, we have 3 possible paths we can take:

1. Add some additional config to the default class
2. Add some point have a `DefaultPlusAgent`, that's a little bit more permissive (in some way the two classes with the UI are doing just that)
3. Create subclass (like a `DefaultAgentXML`, which just replaces the one method) whenever they're needed. So it's up to the user to bind them together in a tiny run script. Or if the class is niche anyway, we might just add it to the cookbook as documentation..

I think I'm kind of on team 3...

klieret:
Hmm, there's also #304, which really points towards XML being maybe the better idea. A very very simple way of supporting both would just be to add a `action_regex` to the config? This would allow maximum flexibility with minimal added complexity

john-b-yang:
Ok any of these three sound great! i don't have a particularly strong opinion. We can go with option three then?

Regarding the other issue - yeah, I'm personally a bigger fan of xml, just because backticks do occasionally pop up in the code.

That makes sense, just declare the regex pattern in the config? I will update the PR to reflect this! I kind of like regex-pattern-in-config, although it might be a bit tricky to specify?

john-b-yang:
Closing this since it seems like we're sticking with backticks now? Feel free to re-open if we re-consider this at some point!

PR #300, opened 2025-07-26
Enh(config): Ask model to look at claude.md & friends

### LINKED ISSUE #292: mini prompt: Ask model to look at instruction prompt files (.cursor, .github/.., .claude etc.)
via @henryiii

klieret:
Closes #292

### DIFF
--- a/src/minisweagent/config/mini.yaml
+++ b/src/minisweagent/config/mini.yaml
@@ -23,7 +23,9 @@ agent:
     You can execute bash commands and edit files to implement the necessary changes.
 
     ## Recommended Workflow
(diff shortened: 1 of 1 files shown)


ai-jz:
@klieret  this is a nice idea!

One question: how does the model read CLAUDE.md? Does it run a command like "vim" to observe its content? However, this will include the whole file content in the context window, right?

klieret:
@ai-jz  It's probably just going to `cat` or similar the file. And yes, if the file is very very long, that's not great (but the same goes for claude code, I assume)

PR #304, opened 2025-07-26
Fix action parser for nested bash ticks

callmephilip:
The current action parser incorrectly parses actions that include nested bash commands in heredocs. [Here's an example](https://github.com/callmephilip/podcast-search/blob/main/history.txt#L297-L453) of this happening in a wild. This is problematic for documentation generation and probably some other spicy use cases.

This is an attempt at fixing this

### DIFF
--- a/src/minisweagent/agents/default.py
+++ b/src/minisweagent/agents/default.py
@@ -104,7 +104,32 @@ class DefaultAgent:
 
     def parse_action(self, response: dict) -> dict:
         """Parse the action from the message. Returns the action."""
-        actions = re.findall(r"```bash\n(.*?)\n```", response["content"], re.DOTALL)
+
+        def extract_top_level_bash_actions(text):
+            """Extract content of top-level bash blocks, avoiding those inside heredocs."""
+
+            # For heredoc blocks: match until EOF, then continue to closing ```
+            # This handles the case where there are nested ```bash blocks inside the heredoc content
+            heredoc_matches = re.findall(r"```bash\n(.*?\nEOF)\n```", text, re.DOTALL)
+
+            if heredoc_matches:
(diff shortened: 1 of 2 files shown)


klieret:
Ah I see, yeah that's a bit problematic.
I appreciate you raising this and adding the PR!

However, my viewpoint is that the current implementation is correct, in that the model is told exactly what it has to expect and should realize itself what the problem is. It can still effectively do these edits in two steps (e.g., first inserting `TRIPLEBACKTICKSbash` and then following up with another `sed` command to turn them into real triple backticks).
Admittedly, that's not the most efficient way.

But I think this issue really points more towards maybe using a custom XML tag that is extremely unlikely to appear outside of this codebase.

callmephilip:
yeah, this sounds reasonable. i'll take a newer version for a spin to see if the new prompt works

doehyunbaek:
Just in case anyone needs it, this is the only code you need to support custom xml based action parsing (with config change of format_error_template). It's a single line change of regex.

```python
class DefaultAgentXML(DefaultAgent):
    def parse_action(self, response: dict) -> dict:
        """Parse the action from the message. Returns the action."""
        actions = re.findall(r"<mini_swe_agent_bash>\n(.*?)\n</mini_swe_agent_bash>", response["content"], re.DOTALL)
        if len(actions) == 1:
            return {"action": actions[0].strip(), **response}
        raise FormatError(self.render_template(self.config.format_error_template, actions=actions))
```

klieret:
Hi @callmephilip @doehyunbaek , thanks for this. As of recently, the regex is configurable and there's an XML config with a regex similar to @doehyunbaek's regex. This should probably solve this issue in almost all cases.

PR #306, opened 2025-07-27
Enh/custom litellm registry

carlosejimenez:
In LitellmModelConfig, we add a litellm_model_registry field.
When set, we update the `litellm` registry with `litellm.register_model`.
This allows us to add models that aren't registered or outdated in the [litellm default registry](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json).


klieret:
This is great! I think I can make it a tiny bit more compact with pathlib

PR #321, opened 2025-07-29
Replace deprecated markdown-link-check with maintained version

0xba1a:
Ref: https://github.com/SWE-agent/SWE-agent/pull/1252


klieret:
Thank you, much appreciated :)

klieret:
Hmm, the new link checker definitely has some other issues. Currently struggling to avoid the following false positives: https://github.com/SWE-agent/mini-swe-agent/actions/runs/16601699375/job/46962863321?pr=323

PR #328, opened 2025-07-31
[WIP] Guardrail via multiagents

ex9-fyi:
Hi, very exciting project - hackable coding agent. I've been thinking about what is the most ambitious version of this. For example, what if `mini` can become the Emacs/[Neo]vim in the coding agent world. I started to wonder how to help agent to make more reliable software. I found the notion of **guardrail* * very interesting - this idea of injecting a message into context to guide the agent at the precise moments the agent needs help e.g. when the agent makes a code change or propose an implementation plan, or when the agent is stuck in loop.

This PR is my attempt to bring the notion of guardrail to `mini`. The first use case is to guide `mini` to develop in TDD style. I would love to get your thoughts on my thinking, approach and implementation, and I am happy to iterate on this PR. I tried to follow the established conventions e.g. `echo` as a return mechanism and use the existing classes like described in [cookbook](https://mini-swe-agent.com/latest/advanced/cookbook/) and keep the implementation minimal. 

# Here is how this PR for multiagents works:

* User i.e. the developer adds a .md file to configure a subagent under `.claude/agents`
* In the system prompt, `mini` tells each agent they have a choice of spawning a subagent and the `description`s of the subagents
* If the agent wants to spawn a subagent, they can `echo MINI_SWE_AGENT_SPAWN_CHILD::<subagent_name>\n<task_description>`. This follows the existing practice of passing data in the `echo` command. `mini` spawn a new agent with fresh context with system prompt and the subagent's configuration loaded into their context.
* Whenever the subagent wants to exit, they exit as in the status quo - `echo MINI_SWE_AGENT_FINAL_OUTPUT\n<data_to_parent_agent>`. The returned data is added to the parent agent's context.
* FYI, at the moment, there is no parallel execution or complex things like that: each agent can choose to synchronously spawn one agent at a time


ex9-fyi:
@klieret

klieret:
Hi @ex9-fyi!

That's very cool! Thanks a lot for experimenting with mini, this is very interesting.

I can see you spent quite some time on this and the implementation looks good, too (though I wonder if cost limits are respected with the sub-agent?)

However, to make it short, I think this might be out of scope for `mini`.

The starting point for `mini` is: "If model providers train models specifically to be agents, what's the simplest agent scaffold we can get away with?".
Sub-agents are just small sub-systems with different context. But if the LM has a long context window and follows the initial instructions well, there is not much of an advantage in having any such sub-system.

Sub-agents also add a lot of complexity (cost limits, output files, how can I browse the history etcetc.), so I don't think it fits quite with what `mini` wants to be.

You could also argue that you can just tell `mini` to call `mini` (the problem would be the timeouts and the cost tracking though), and I think that would fit slightly better with the core hypothesis here: "code little, delegate as much as possible to the LM".

However, what I could offer is to create a space/showcase in the documentation where we link to forks and ideas. What do you think?

src/minisweagent/agents/default.py
```diff
@@ -72,7 +72,7 @@ def add_message(self, role: str, content: str):
     def run(self, task: str) -> tuple[str, str]:
         """Run step() until agent is finished. Return exit status & message"""
         self.messages = []
-        self.add_message("system", self.config.system_template)
+        self.add_message("system", self.render_template(self.config.system_template))
```
klieret, line 75: Happy to merge a PR with this change!

src/minisweagent/agents/interactive.py
```diff
@@ -35,8 +35,8 @@ class InteractiveAgentConfig(AgentConfig):
 class InteractiveAgent(DefaultAgent):
     _MODE_COMMANDS_MAPPING = {"/u": "human", "/c": "confirm", "/y": "yolo"}
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, config_class=InteractiveAgentConfig, **kwargs)
+    def __init__(self, *args, config_class=InteractiveAgentConfig, **kwargs):
```
klieret, line 38: Also happy to add this in a separate PR. Maybe add a `*` before `config_class` to mark it keyword-only

ex9-fyi:
"""
The starting point for mini is: "If model providers train models specifically to be agents, what's the simplest agent scaffold we can get away with?".
Sub-agents are just small sub-systems with different context. But if the LM has a long context window and follows the initial instructions well, there is not much of an advantage in having any such sub-system.

Sub-agents also add a lot of complexity (cost limits, output files, how can I browse the history etcetc.), so I don't think it fits quite with what mini wants to be.
"""

I see where you're coming from. : )

I'd love to try out a multi-agent scaffold on SWE verified for the purpose of "guiding" the agent and keep you posted on the performance. At this point, I don't know if it helps but my intuition based off of experience using Claude Code: it feels there is more hallucination after context window is filled half-way or more. Let me know if you know something tried conserving context / use TDD to increase performance.

Happy to revisit this PR if I find gains :).

On refactoring interactive.py and default.py, and a "new idea section", here we go, glad you find these interesting: 
* https://github.com/SWE-agent/mini-swe-agent/pull/339/files
* https://github.com/SWE-agent/SWE-agent/pull/1263/files

ex9-fyi:
I realize this PR could use an example to see how subagent works in practice: https://gist.github.com/ex9-fyi/202fda447a6e59bcf6d6466f5427ac79

PR #381, opened 2025-08-08
Change: exit sequence; SWE-bench prompts

klieret:
- **tmp**
- **Change: exit sequence; SWE-bench prompts**


closji:
Very minor, but if you're not majorly opposed, I still think `COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT` is a slightly better working for the final keyword.

Issue #386, opened 2025-08-08
Running swebench with singularity / apptainer


pcmoritz:
### Describe the issue

Thanks for making this project, I really like the design decisions you made (especially around stateless execution).

I was trying to run swebench with singularity / apptainer but was running into the following problem -- I patched the code like this:
```patch
diff --git a/src/minisweagent/environments/singularity.py b/src/minisweagent/environments/singularity.py
index b9fa27d..c709b7e 100644
--- a/src/minisweagent/environments/singularity.py
+++ b/src/minisweagent/environments/singularity.py
@@ -27,7 +27,7 @@ class SingularityEnvironment:
 
     def execute(self, command: str, cwd: str = "") -> dict[str, Any]:
         """Execute a command in a Singularity container and return the result as a dict."""
-        cmd = [self.config.executable, "exec"]
+        cmd = [self.config.executable, "exec", "--writable"]
 
         work_dir = cwd or self.config.cwd
         if work_dir and work_dir != "/":
diff --git a/src/minisweagent/run/extra/swebench_single.py b/src/minisweagent/run/extra/swebench_single.py
index eadf219..869ca1c 100644
--- a/src/minisweagent/run/extra/swebench_single.py
+++ b/src/minisweagent/run/extra/swebench_single.py
@@ -9,6 +9,7 @@ from datasets import load_dataset
 from minisweagent.agents.interactive import InteractiveAgent
 from minisweagent.config import builtin_config_dir, get_config_path
 from minisweagent.environments.docker import DockerEnvironment
+from minisweagent.environments.singularity import SingularityEnvironment
 from minisweagent.models import get_model
 from minisweagent.run.extra.swebench import DATASET_MAPPING, get_swebench_docker_image_name
 
@@ -40,7 +41,7 @@ def main(
     instance: dict = instances[instance_spec]  # type: ignore
 
     _config = yaml.safe_load(get_config_path(config_path).read_text())
-    env = DockerEnvironment(**(_config.get("environment", {}) | {"image": get_swebench_docker_image_name(instance)}))
+    env = SingularityEnvironment(**(_config.get("environment", {}) | {"image": "docker://" + get_swebench_docker_image_name(instance)}))
     agent = InteractiveAgent(
         get_model(model_name, _config.get("model", {})),
         env,
```

but I'm getting the following error:
```
FATAL:   no SIF writable overlay partition found in <directory>
```

[...]

PR #387, opened 2025-08-08
Make it possible to run swebench in the singularity environment

pcmoritz:
Fixes https://github.com/SWE-agent/mini-swe-agent/issues/386

Since singularity containers are read only by default, this creates a sandbox https://apptainer.org/docs/user/main/cli/apptainer_build.html in which the commands are executed.

### DIFF
--- a/docs/blog/posts/gpt5.md
+++ b/docs/blog/posts/gpt5.md
@@ -127,7 +127,7 @@ So the real winner in my opinion is `GPT-5-mini`!
     Check out our [notes](../../quickstart.md#gpt-5) on how to run `mini` with GPT-5.
 
     You can reproduce our numbers in this blog by following the [swebench](../../usage/swebench.md) tutorial, but
-    the tl;dr is to run (remove the temperature setting from the `swebench.yaml` file first, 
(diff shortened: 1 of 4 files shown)


ofirpress:
Thanks for the PR!

pcmoritz:
Using this code, I was able to successfully run swebench via `mini-extra swebench --model gpt-5-mini --subset verified --split test -e singularity --workers 8` :)

klieret:
This is great, really appreciate it @pcmoritz !

I'm gonna merge this in, then probably push a small refactor on top and move `get_environment` to `environments/__init__`, since I can see this coming up more often, anyway (and this would mirror the structure that we use for the model classes in `models/__init__`.

I'll also check if I can extend the tests a bit (seeing that this didn't work before, some of the desired behavior clearly wasn't covered by the tests)

pcmoritz:
Thanks @klieret that would be amazing! I also saw there were some other scripts that are currently hard-coded to docker where using the `get_environment` function would be helpful, so this refactor would be very much appreciated!

klieret:
(The test failures are probably trivial (need to specify all args to `main` because `typer` nonsense))

PR #388, opened 2025-08-09
feat: Graceful handling of unknown models in LitellmModel

psukhopompos:
## Summary
This PR makes mini-swe-agent work seamlessly with any OpenRouter model (or other providers) even when the model isn't in litellm's pricing registry. Instead of crashing, it continues working and provides helpful guidance.

## Problem
Currently, when using models not in litellm's registry (like many OpenRouter models), mini-swe-agent crashes with:
```
ValueError: This model isn't mapped yet. Add it here - https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json
```

This is frustrating for users who just want to try a new model quickly.

## Solution
- Wrap cost calculation in try/except
- Continue with cost=0 if model is unknown  
- Show a helpful warning (once per model) explaining how to enable cost tracking
- Track how many calls were made without cost info

## Benefits
1. **Just works** - Any model from any provider works immediately
2. **Helpful** - Clear instructions for users who want cost tracking
3. **Minimal** - ~20 lines, fits mini-swe-agent's philosophy
4. **Non-intrusive** - Only warns once per model, not every call

## Testing
Added comprehensive tests in `test_unknown_model_handling.py`:
- ✅ Unknown models work without registry file
- ✅ Shows helpful warning on first use only
- ✅ Known models still track costs properly
- ✅ Tracks unknown_model_calls metric

Tested with real OpenRouter kimi-k2 model.

## Example Output
```
⚠️  Cost tracking unavailable for 'openrouter/moonshotai/kimi-k2'
   To enable cost tracking, create a model registry file:
   1. Create model_registry.json with your model's pricing
   2. Set LITELLM_MODEL_REGISTRY_PATH=/path/to/model_registry.json
   See: https://docs.litellm.ai/docs/proxy/custom_pricing
```

## Related Issues
Addresses the need for "Support for more models (anything where litellm doesn't work out of the box)" mentioned in contributing.md.

This aligns with mini-swe-agent's philosophy: minimalistic, hackable, and practical.

### DIFF
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -18,6 +18,9 @@ from minisweagent.models import GLOBAL_MODEL_STATS
 
 logger = logging.getLogger("litellm_model")
 
+# Track which models we've already warned about
+_warned_models = set()
+
 
 @dataclass
 class LitellmModelConfig:
@@ -31,6 +34,7 @@ class LitellmModel:
         self.config = LitellmModelConfig(**kwargs)
         self.cost = 0.0
         self.n_calls = 0
+        self.unknown_model_calls = 0
         if self.config.litellm_model_registry and Path(self.config.litellm_model_registry).is_file():
             litellm.utils.register_model(json.loads(Path(self.config.litellm_model_registry).read_text()))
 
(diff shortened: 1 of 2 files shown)


klieret:
Hi @psukhopompos thanks for taking the time for this! 

However, costs are a dangerous thing, overrunning spending limits is probably the most dangerous thing an application can let happen. I can see use cases where users use mini in CI, or in other scenarios where output might not be read immediately. And then, they try a new model one day, and suddenly it runs for forever, spending huge amounts of money.

I don't think we should leave any chance to this. However, I think adding a helpful message in case of error is helpful. Something super trivial like

```
try:
     calc cost
except Exception as e:
     raise RuntimeError(f"Issue during cost calculation. See 'Updating the model registry' at https://mini-swe-agent.com/latest/advanced/local_models/ if you need to add a model.")
```

or something like that. 

Happy to merge a PR for that or I'll create one in the next two days or so :)

PR #402, opened 2025-08-12
Ref: get_environment

klieret:


klieret:
@pcmoritz I've done some refactoring (and will bring this to the other run scripts shortly). Will follow up with something similar also for the agent and model class (probably).

To make this more standardized, the new flag is now `--environment-type`

Again, thanks for getting this started!

PR #411, opened 2025-08-13
change swe-rex env usage to set merge_output_streams True

carlosejimenez:
This change depends on updates introduced in swe-agent/swe-rex#238 .
Upon merging that PR, we should merge this one, and set a new minimum swe-rex version in the requirements.

This sets `merge_output_streams` to True when invoking RexCommand for the swerex_docker environment. Currently, the non-multiplexed format is incompatible with `has_finished` disallowing agents to actually complete tasks in swe-bench.


klieret:
Yeah was thinking about that too! We're really due a new swe-rex release 😅 
I'm gonna try and do this today

closji:
Maybe pin swerex>=1.4.0 and merge?

klieret:
Will do, one second!

PR #425, opened 2025-08-14
Fix: Get env template vars from environment itself

### LINKED ISSUE #419: When running with docker, templating environment info is taken from host system, not container
Way to solve: Get env from environment via method

klieret:
> [!warning]
> This PR has been mostly rolled back in #429 

Closes #419

### DIFF
--- a/src/minisweagent/environments/docker.py
+++ b/src/minisweagent/environments/docker.py
@@ -2,7 +2,7 @@ import os
 import shlex
 import subprocess
 import uuid
-from dataclasses import dataclass, field
+from dataclasses import asdict, dataclass, field
 from typing import Any
 
 from minisweagent.environments.utils.template_vars import get_remote_template_vars
@@ -40,7 +40,7 @@ class DockerEnvironment:
         self._start_container()
 
     def get_template_vars(self) -> dict[str, Any]:
-        return get_remote_template_vars(self)
+        return asdict(self.config) | get_remote_template_vars(self)
 
     def _start_container(self):
         """Start the Docker container and return the container ID."""
--- a/src/minisweagent/environments/extra/swerex_docker.py
+++ b/src/minisweagent/environments/extra/swerex_docker.py
@@ -1,10 +1,11 @@
 import asyncio
-import json
 from dataclasses import asdict, dataclass, field
 from typing import Any
 
 from swerex.deployment.docker import DockerDeployment
 from swerex.runtime.abstract import Command as RexCommand
+
+from minisweagent.environments.utils.template_vars import get_remote_template_vars
 
 
 @dataclass
@@ -45,10 +46,4 @@ class SwerexDockerEnvironment:
         }
 
     def get_template_vars(self) -> dict[str, Any]:
-        try:
-            platform_info = json.loads(
-                self.execute("python -c 'import platform; print(platform.uname()._asdict())'")["output"]
-            )
-        except ValueError:
-            platform_info = {}
-        return platform_info | asdict(self.config)
+        return asdict(self.config) | get_remote_template_vars(self)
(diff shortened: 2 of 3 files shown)


src/minisweagent/environments/utils/template_vars.py
```diff
@@ -0,0 +1,41 @@
...
+_PLATFORM_INFO_COMMAND = """python -c '
+import platform;
+import json
+print(json.dumps(platform.uname()._asdict()))
+'
+"""
+_ENV_COMMAND = "env"
+
+
+def get_remote_template_vars(env: Environment, strict: bool = False) -> dict[str, Any]:
```
pcmoritz, line 17: I'm curious why you need to do something like this, it seems pretty brittle to me (e.g. even assuming python is in the container is probably not a good assumption).

Shouldn't it always work like this: If any extra env variables are specified in the config, they are added to the environment by augmenting the command line of docker or singularity. Otherwise we don't make any changes about the environment and don't try to get the environment of the container. With this mindset, the proper fix to https://github.com/SWE-agent/mini-swe-agent/issues/419 would be to just remove the `os.environ` part.

it is quite possible that I don't understand which problem you are trying to solve here though, so take what I'm saying with a grain of salt. If I understand the `render_template` part correctly, I would expect the llm to just run `env` in their environment if it needs environment variables :)

klieret: Hi @pcmoritz thanks for taking a look :) 

You're right, it's probably a little bit brittle, but also nothing really rests on this. 

The motivation came from the current

```
  {%- if system == "Darwin" -%}
    <important>
    You are on MacOS. For all the below examples, you need to use `sed -i ''` instead of `sed -i`.
    </important>
    {%- endif -%}
```

bit in the `mini.yaml` config (this is not part of the `swebench.yaml`). I noticed that this is going to be misleading if we run it on docker.

But you're right, another way to address this is to just not give any env equivalent. 

I think either way, I wanted to move to the `get_template_vars()` method though because it allows the model and environment class (and subclasses thereof) a little bit more freedom to include extra information that possibly changes over the course of the run (for another project we're running mini with very limited maximum step count and we want to tell the model how many steps are left to take, for example).

klieret: But yeah, let me think about the platform bit specifically. 

If we want to keep some basic platform information for the prompts, we could also just check for `if Darwin in system_info_str` in the jinja template and then try a few different ways to get it in 
`get_template_vars` (like `uname -a || python -c ... || whatever` and then we just take this as a string).

klieret: OK I also realized that the currrent implementation is a little bit silly, because we do the `.execute` at every step, even though environment variables can't even be changed

klieret: OK, thought some more and you're right, let's get rid of this for the most part, it's just not worth this kind of niche use case.

pcmoritz: Got it, that makes sense -- I think if you want to do something like this (expose environment info to the template), a very natural way to do it would be to make it user configurable on which command gets run in the environment to get that information, and then expose the output of the command to the template. That would be most in line with the philosophy of minimalism I think. Users who don't want to do something like this could just turn it off.

The sed fix seems a little specific, I wonder if the LLM could find out sed behaves differently itself. It is also not specific to macOS but will most likely be the same on other BSD based operating systems.

klieret: Agreed with what you wrote!

> The sed fix seems a little specific, I wonder if the LLM could find out sed behaves differently itself. It is also not specific to macOS but will most likely be the same on other BSD based operating systems.

Agreed, and the model might find out itself, but also this is one of the compromises with `mini` just working reasonably as a command line helper (you just don't want to wait for it to find these things out every single time). And a lot of the users are probably on macos, so this should work.

klieret: (it's not part of the benchmark configuration)

Issue #401, opened 2025-08-11
Lightweight sandboxing through wrapper commands


pcmoritz:
### Describe the issue

I've recently been experimenting with some lightweight sandboxing strategies for SWE like tasks that don't require full docker images but still allow for sandboxing and environment setup (see e.g. https://github.com/NovaSky-AI/SkyRL/pull/80 and https://github.com/NovaSky-AI/SkyRL/pull/129). The advantage of that are of course that we don't need to deal with building docker images, docker daemons or docker rate limits.

Some examples include:

- Just running the commands in a virtual environment set up with uv
- Running the commands in a https://github.com/containers/bubblewrap environment (in which case the command needs to be prefixed by an appropriate `bwrap` command)
- Running the commands in an environment set up with e.g. `guix shell` (see https://dthompson.us/posts/guix-for-development.html) or `nix-shell`

I believe these could all be supported well by mini-swe-agent if we make a small extension to the local environment https://github.com/SWE-agent/mini-swe-agent/blob/main/src/minisweagent/environments/local.py -- we would need to add:

- An optional `wrapper_cmd` that gets prepended to the command in the subprocess.run command, this would be the `guix shell` or `bwrap` wrapper
- An optional `setup_cmd` that will be used to setup the environment (e.g. uv venv + uv pip install)
- An optional `cleanup_cmd` that will be run at the end of the command execution in the `__del__ `method of the environment and would be used to clean up the environment if any was set up.

It is possible that the other environments (singularity and docker) could actually also fit into this abstraction, but I haven't thought too deeply about whether that is a good idea.

Curious about your thoughts. It seems to me that this would fit well with the philosophy of this project, but could also perceived as too much (in which case such an environment could be maintained as an extra or as an example if there is sufficient interest).

klieret:
Very supportive of that!! 

And I agree that it looks like we could do that very easily, especially now that we also have the `get_environment` switch in there.

I see a few different options for implementations:

1. Extend `LocalEnvironment` with the abovementioned settings
   * Pro: Low amount of duplicate code
   * Con: (1) Might be slippery slope to end up with a bloated class (e.g., don't we need to generate names for the venvs?); `local.py` is intended to be one of the very slim classes (2) requires very specific config file changes to switch environment
2. Create `WrappedLocalEnvironment` or a similar "more advanced class"
   * Pro: flexible
   * Con: Requires very specific config file changes to switch environment, making it somewhat hard to use
[...]

PR #452, opened 2025-08-19
Add bubblewrap environment

pcmoritz:
This is a minimal version to support https://github.com/SWE-agent/mini-swe-agent/issues/401

Bubblewrap supports lightweight sandboxing without the need to use container images. It just mounts a subset of the host file system into the sandbox. The existing class can easily be extended to e.g. set up a virtual environment with uv, or clone a github repository by calling `execute` after the environment is instantiated.


pcmoritz:
@klieret This is now ready for review :)

klieret:
Super cool, thank you so much for the work! Might need till tomorrow to take a look, today was a bit scattered!

pcmoritz:
I've now switched over to this environment for most of my swebench like runs, and this is how I am integrating it into the `swebench.py` script at the moment (slightly hacky since `get_swebench_docker_image_name` should not be called if the bubblewrap environment is used):

```patch
diff --git a/src/minisweagent/run/extra/swebench.py b/src/minisweagent/run/extra/swebench.py
index 30cbb31..1ca3612 100644
--- a/src/minisweagent/run/extra/swebench.py
+++ b/src/minisweagent/run/extra/swebench.py
@@ -79,6 +79,14 @@ def get_swebench_docker_image_name(instance: dict) -> str:
 def get_sb_environment(config: dict, instance: dict) -> Environment:
     image_name = get_swebench_docker_image_name(instance)
     env_config = config.get("environment", {})
+    if env_config.get("environment_class") == "bubblewrap":
+        env = get_environment(env_config)
+        result = env.execute(
+            f"git init && git remote add origin https://github.com/{instance['repo']} && "
+            f"git fetch --depth 1 origin {instance['base_commit']} && git checkout FETCH_HEAD && cd ..",
+        )
+        assert result["returncode"] == 0, result["output"]
+        return env
     if env_config.get("environment_class") == "singularity":
         image_name = "docker://" + image_name
     env_config["image"] = image_name
```

If there is any way you see to do this better or integrate it into the current code I'd love to learn about it :)

klieret:
Hi Moritz! Sorry it took me a while to take a look at this! Let me get this merged, then make sure we have bubblewrap installed on the CI, so it properly tests.

klieret:
my todos

* [x] Add env var to docs
* [x] Get testing with CI
* [x] Add to `get_environment` (?)
* [x] Include env in docs

klieret:
Btw, looking at your integration piece of code, the `cd ..` probably doesn't do anything (?)  since it's still a subshell

klieret:
Hmm, I guess one way of doing this elegantly could be to have the ability to add a startup command (which could get templated with the instance variables). We could add this to a new `extra` (or `run` or `swebench`) key in the config.

pcmoritz:
Thanks for merging the PR!

Ah you are right, the `cd ..` doesn't do anything, it was left over from a previous place I used this code in :)

Making it possible to add a startup command like this makes sense to me, at the same time we probably also don't want people to over use it. Which config do you think it should be added to?

My first name is Philipp btw (my last name is Moritz) 😁

klieret:
> My first name is Philipp btw (my last name is Moritz) 😁

Ahaha, sorry got tripped up 🙏   Appreciate the correction!

klieret:
Btw, I'm getting some issues on CI https://github.com/SWE-agent/mini-swe-agent/pull/463

```
test_bubblewrap_environment_multiline_output result: {'output': 'bwrap: setting up uid map: Permission denied\n', 'returncode': 1}
```

I can try on a linux machine later. Probably have to read up a biton what's up with the setuid

pcmoritz:
Thanks, I didn't have that problem when I was running it (maybe my ubuntu is configured differently than the CI), let's try to disable the user isolation (it is fine to run as the same user that executed the `bwrap` executable). There seems to be a `--disable-userns` option, let's try to add that to the default args for bwrap and see if that makes it run on the CI? If that works, it will make it more likely to run in other environments that users might use.

pcmoritz:
Do you know which version of bwrap is used in the CI? It seems like `--disable-userns` was only introduced in 0.8.0: https://github.com/containers/bubblewrap/pull/488 and https://github.com/containers/bubblewrap/releases/tag/v0.8.0

I was using 0.6.1 which doesn't seem to have this flag yet. Ideally we figure out a way that makes this work on both (all) versions.

klieret:
It seems like at least in ubuntu 0.8.0 is available since noble numbat (last april), so yeah still somewhat new... But let me first try if this fixes it

klieret:
Hmm, that's weird, it's the same error still.
(also apparently you need to add `"--unshare-user" for "--disable-userns" to even be accepted)

klieret:
I wonder if we could use `--userns` and directly set it to the current user namespace or something

klieret:
OK I think I got it. It was a combination of `--unserhare-user-try` and adding the setuid bit to `bubblewrap` in the setup

pcmoritz:
Got it, something else we could try is not set up a new userns, e.g. by explicitly specifying `--uid` and `--gid` of the current user (e.g. in my case that is `--uid 1000 --gid 100`). That might be preferable since it will work with all versions of bubblewrap and probably without setuid too.

klieret:
Seems like `--unshare-user-try` has been with it from v0.1.8, so I think we're good!

pcmoritz:
Ah you are right, sounds good to me!

klieret:
Regarding the swebench integration, I think we should add a `run: env_startup_command: str` field in the swebench config (i.e., `swebench.yaml`) that's templated with the swe-bench instance variables. That way we could put in specific behavior. I think this might also be required here: https://github.com/SWE-bench/SWE-smith/issues/147

And this would be relatively unobtrusive & works also for all folks who run on swebench-like datasets but with some special setup

klieret:
Let's continue in #468

PR #487, opened 2025-08-29
[SWEBench] Add support for SWEBench evaluation

SumanthRH:
# What does this PR do?

WIP PR that adds support for evaluation for SWEBench in `mini-extra`. Currently only supported for batched version and for the apptainer environment 

TODO: 
- [x] Verify correctness by reproducing GPT-5-nano numbers. 
- [x] Cleanup

### DIFF
--- a/src/minisweagent/environments/singularity.py
+++ b/src/minisweagent/environments/singularity.py
@@ -25,6 +25,8 @@ class SingularityEnvironmentConfig:
     """Path to the singularity executable."""
     sandbox_build_retries: int = 3
     """Number of retries for building the sandbox if an error occurs."""
+    writeable_tmp: bool = False 
+    """Allow for a writeable /tmp directory in the sandbox. This is helpful for adding evaluation scripts outside of the working directory"""
 
 
 class SingularityEnvironment:
@@ -66,6 +68,9 @@ class SingularityEnvironment:
 
         # Do not inherit directories and env vars from host
         cmd.extend(["--contain", "--cleanenv"])
+
+        if self.config.writeable_tmp:
+            cmd.extend(["--no-mount", "tmp"])
 
         work_dir = cwd or self.config.cwd
         if work_dir and work_dir != "/":
--- a/src/minisweagent/environments/singularity.py
+++ b/src/minisweagent/environments/singularity.py
@@ -94,6 +94,15 @@ class SingularityEnvironment:
         )
         return {"output": result.stdout, "returncode": result.returncode}
 
+    def copy_to(self, local_filepath: str, dst_filepath: str):
+        cmd = ["cp", local_filepath, str(self.sandbox_dir / dst_filepath.lstrip("/"))]
+        result = subprocess.run(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+        )
+        return {"output": result.stdout, "returncode": result.returncode}
+
     def cleanup(self):
         shutil.rmtree(self.sandbox_dir, ignore_errors=True)
 
--- a/src/minisweagent/environments/singularity.py
+++ b/src/minisweagent/environments/singularity.py
@@ -25,7 +25,7 @@ class SingularityEnvironmentConfig:
     """Path to the singularity executable."""
     sandbox_build_retries: int = 3
     """Number of retries for building the sandbox if an error occurs."""
-    writeable_tmp: bool = False 
+    writeable_tmp: bool = False
     """Allow for a writeable /tmp directory in the sandbox. This is helpful for adding evaluation scripts outside of the working directory"""
 
 
--- a/src/minisweagent/run/extra/swebench.py
+++ b/src/minisweagent/run/extra/swebench.py
@@ -17,12 +17,11 @@ import yaml
 from datasets import load_dataset
 from rich.live import Live
 
-from minisweagent import Environment
 from minisweagent.agents.default import DefaultAgent
 from minisweagent.config import builtin_config_dir, get_config_path
-from minisweagent.environments import get_environment
 from minisweagent.models import get_model
 from minisweagent.run.extra.utils.batch_progress import RunBatchProgressManager
+from minisweagent.run.extra.utils.swebench_utils import get_sb_environment
 from minisweagent.run.utils.save import save_traj
 from minisweagent.utils.log import add_file_handler, logger
 
@@ -63,26 +62,6 @@ class ProgressTrackingAgent(DefaultAgent):
             self.instance_id, f"Step {self.model.n_calls + 1:3d} (${self.model.cost:.2f})"
         )
         return super().step()
-
-
-def get_swebench_docker_image_name(instance: dict) -> str:
-    """Get the image name for a SWEBench instance."""
-    image_name = instance.get("image_name", None)
-    if image_name is None:
-        # Docker doesn't allow double underscore, so we replace them with a magic token
-        iid = instance["instance_id"]
-        id_docker_compatible = iid.replace("__", "_1776_")
-        image_name = f"swebench/sweb.eval.x86_64.{id_docker_compatible}:latest".lower()
-    return image_name
-
-
-def get_sb_environment(config: dict, instance: dict) -> Environment:
-    image_name = get_swebench_docker_image_name(instance)
-    env_config = config.setdefault("environment", {})
-    if env_config.get("environment_class") == "singularity":
-        image_name = "docker://" + image_name
-    env_config["image"] = image_name
-    return get_environment(env_config, default_type="docker")
 
 
 def update_preds_file(output_path: Path, instance_id: str, model_name: str, result: str):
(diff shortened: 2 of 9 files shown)


klieret:
Hi, thanks for working on this @SumanthRH! But I think if we integrate automatic SB evaluations, we might want to go with https://www.swebench.com/sb-cli/ as it's much easier and faster (though it only supports the "official" datasets).

Alternatively, if you want to run with some custom evaluation harness, one more thing we could do is to add support for a hook in some way? So that you could easily integrate any eval script that runs after every finished instance? 

I'm just a bit wary of introducing a quite non-trivial and quite specific piece of code here like this. 

Could you tell me more about the usecase?

pcmoritz:
@klieret It certainly won't make sense to merge the code in its current form, but I wonder if it makes sense to have a minimal evaluation setup as part of this repo. The main benefit over other evaluation setups would be that it would support all the backends that mini-swe-agent supports (like apptainer). We will need some code like that to support RL, since there the eval needs to run as part of the RL loop.

If we want to have first class support for mini-swe-agent in https://github.com/NovaSky-AI/SkyRL, we will need eval code that is flexible with regard to the environment class the agent uses. We could also develop that in the SkyRL repo, but if you are open to it, it would be great to put the code into the mini-swe-agent repo, since I think it would also be very valuable outside of the RL scenario.

If you like we can come up with a minimal design to support this together.

pcmoritz:
I'll leave some review comments on this PR concerning what I think should be done differently if we were indeed going to merge some code that allows doing evals.

src/minisweagent/run/extra/swebench.py
```diff
@@ -110,11 +89,28 @@ def remove_from_preds_file(output_path: Path, instance_id: str):
...
+        eval_result["evaluation_error"] = e
+    return eval_result
+
+
 def process_instance(
     instance: dict,
     output_dir: Path,
     config: dict,
     progress_manager: RunBatchProgressManager,
+    run_eval: bool = False,
```
pcmoritz: I think it would be better to not integrate the eval code into the existing `swebench.py` script, and instead make a separate script that only does evals given the output of the `swebench.py` file, so the script stay simple.

src/minisweagent/run/extra/swebench_eval.py
```diff
@@ -0,0 +1,187 @@
...
+    if not isinstance(patch, str):
+        return ""
+
+    if not patch.strip():
+        # skip empty patches
+        return ""
+
+    patch = patch.replace("\r\n", "\n")
+    # There might be some weird characters at the beginning of the patch
+    # due to some OpenHands inference command outputs
```
pcmoritz: We should not have compatibility shims like this (this whole `process_git_patch` function should not exist). Instead just assume the patch is produced by `mini-swe-agent` in a clean format.

src/minisweagent/environments/singularity.py
```diff
@@ -25,6 +25,8 @@ class SingularityEnvironmentConfig:
     """Path to the singularity executable."""
     sandbox_build_retries: int = 3
     """Number of retries for building the sandbox if an error occurs."""
+    writeable_tmp: bool = False
```
pcmoritz, line 28: Why is this needed? Instead of writing the eval script to a file shared with the host, can we just run it via `.execute`?

src/minisweagent/run/extra/swebench_eval.py
```diff
@@ -0,0 +1,187 @@
...
+    obs = env.execute("chmod +x /tmp/eval.sh", cwd="/")
+    assert obs["returncode"] == 0, (
+        f"Got bad observation: {obs} while executing `chmod +x /tmp/eval.sh` for environment dir: {env.sandbox_dir}"
+    )
+
+    # Apply patch
+    # TODO (sumanthrh): This is from SkyAgent: https://github.com/NovaSky-AI/SkyRL/tree/main/skyagent, should probably just match the eval function in SWE-Bench repo
+    exec_command = (
+        "cd /testbed && "
+        "(git apply -v /tmp/patch.diff && echo 'APPLY_PATCH_PASS' || "
```
pcmoritz: We should probably be a lot more opinionated here, only support patches that apply cleanly with `git apply`, execute it with `.execute` and check if the return code is 0 to see if it was successful.

I realize this will not be compatible with the swebench eval and we should not claim it is. It will just be an eval that can evaluate mini-swe-agent traces well and be integrated into an RL setup. The final eval on the quality of the policy would still need to be done with the swebench harness if exact compatibility is desired.

src/minisweagent/run/extra/swebench_eval.py
```diff
@@ -0,0 +1,187 @@
...
+        raise Exception(f"Instance {instance_id} {APPLY_PATCH_FAIL}:\n{apply_patch_output}")
+    elif "APPLY_PATCH_PASS" in apply_patch_output:
+        logger.info(f"[{instance_id}] {APPLY_PATCH_PASS}:\n{apply_patch_output}")
+
+        # Run eval script in background and save output to log file
+        log_file = "/tmp/eval_output.log"
+        command = f"/tmp/eval.sh > {log_file} 2>&1 & echo $!"
+        obs = env.execute(command, cwd="/")
+
+        if isinstance(obs, dict) and obs["returncode"] == 0:
```
pcmoritz: Hmm, this whole code is not very mini-swe-agent-esque, instead it would be better to just support a user defined eval script (which would e.g. be the command to run the test suite of the project), which gets run with `.execute`, and then support a simple filter on the stdout to check if the tests passed or not. That filter could come in the form of a python function that the user can plug in (with a swebench like one as an example), or maybe a regex or similar is expressive enough, I'm not really sure (but the regex is an instance of a python function, so it fits into that pattern).

The problem with a python function is it is not super easy to plug one in by the user, we would need to think a bunch more about that. A function could be provided as a string and then `eval`ed or a python file with the code could be imported at runtime but these both don't seem super satisfactory.

src/minisweagent/run/extra/swebench_eval.py
SumanthRH: Makes sense! Still cleaning this up!

ofirpress:
i haven't had a time to look deeply at any of this so take my input with a grain of salt but @klieret don't you think having sb evals here would be helpful? how do you do evals now for swebench.com? 

also in general i know that from a user's perspective sb-cli is simplest, but we [the swe-bench team] have to spend our budget on maintaining that, do we really want to push more usage of sb-cli? wouldn't it be better if everyone could run sb evals client side? (at least for ML and V which are public)

pcmoritz:
I'd suggest to have a very simple eval framework in `mini-swe-agent`:

Allow having one shell script per problem instance (which can be part of the dataset), which takes the patch generated by the LLM as an argument, and will perform the eval -- if it returns return code 0, that indicates "success" and any other return code indicates "failure".

I think the existing swe-bench evals would fit into this, you can just make the eval script generated by `make_test_spec` that script. Other evals (like running the full test suite of the "golden" PR) can be put into this framework and it is pretty flexible overall. And it removes a bunch of complexity from the code and puts it into the dataset, which makes it easier for future datasets to evolve it vs. being too tied to the concrete swebench logic. It is dead simple and fits the philosophy of `mini-swe-agent` pretty well I think, the script can just be executed via the `execute` function.

Maybe a slight variant of this should be used (e.g. instead of a shell script, it could be a python script). But it feels pretty nice to me, curious what others think :)

SumanthRH:
Hi @klieret . I'm currently refactoring this PR after offline discussion with @pcmoritz to implement his suggestion in the above comment: https://github.com/SWE-agent/mini-swe-agent/pull/487#issuecomment-3238942503

I have a basic version right now, will cleanup and let you know when this is ready

closji:
@pcmoritz , the currently supported local method for evaluating predictions on SWE-bench is to use https://github.com/swe-bench/swe-bench

Does this not currently fit your needs?

pcmoritz:
@carlosejimenez Thanks for pointing that out, I think it has two shortcomings:

1. It only supports docker and not the other environments offered by `mini-swe-agent`
2. It is very tied to the swe-bench dataset / setup, ideally we would have something that is more general and allows more flexible evaluations (and is also simpler)

closji:
We're actively developing an update to the SWE-bench repository that will make running with arbitrary images and datasets much easier. We'll try our best to push this out as soon as possible.
I hope that that update will resolve some of your needs.
Regarding this PR, I think the best solution for you is to simply maintain these scripts on a separate dev fork for the time being.
Alternatively, I think @klieret 's idea to incorporate a hook to make running a script automatically upon completion could be an acceptable short-term compromise.

pcmoritz:
Thanks sounds good! There is a new version in https://github.com/SWE-agent/mini-swe-agent/pull/501 of this PR which is much simpler and implements the separate eval script (together with a dataset that has the required format https://github.com/SWE-agent/mini-swe-agent/pull/501/files#diff-aed655de6b25e314e855e68e7b9473259c8803b77d54f0a498c6bcbc71513e7fR110), for the time being I'll probably just use that and keep iterating on it in that PR, it should be easy to keep it in sync with the main branch :)

ofirpress:
@pcmoritz apologies if my post above mislead you, i should've waited until the weekly swe-bench sync before responding.

anyways, we're looking forward to hearing your thoughts after we release swe-bench.next, and are very open to making more modifications if it's not good enough then. 

our aim is for mini-swe-agent to be the quickest + simplest way to evaluate models on swe-bench.

PR #490, opened 2025-08-29
Feat: Support for SWE-bench startup command

### LINKED ISSUE #468: Allow to execute container startup command in swebench/swebench-single
See discussion in https://github.com/SWE-bench/SWE-smith/issues/147 and https://github.com/SWE-agent/mini-swe-agent/pull/452#issuecomment-3211905425 (https://github.com/SWE-agent/mini-swe-agent/pull/452)

@pcmoritz

klieret:
Closes #468

### DIFF
--- a/src/minisweagent/run/extra/swebench.py
+++ b/src/minisweagent/run/extra/swebench.py
@@ -15,6 +15,7 @@ from pathlib import Path
 import typer
 import yaml
 from datasets import load_dataset
+from jinja2 import Template
 from rich.live import Live
 
 from minisweagent import Environment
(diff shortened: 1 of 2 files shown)


src/minisweagent/run/extra/swebench.py
```diff
@@ -77,12 +78,18 @@ def get_swebench_docker_image_name(instance: dict) -> str:
...
+    env_config["environment_class"] = env_config.get("environment_class", "docker")
+    image_name = get_swebench_docker_image_name(instance)
+    if env_config["environment_class"] == "docker":
+        env_config["image"] = image_name
+    elif env_config["environment_class"] == "singularity":
+        env_config["image"] = "docker://" + image_name
+    env = get_environment(env_config)
+    if startup_command := config.get("run", {}).get("env_startup_command"):
+        startup_command = Template(startup_command).render(**instance)
+        env.execute(startup_command)
```
pcmoritz: Can we check that the return code was zero here and raise an exception if it is not? The rationale is: If the setup failed, I really want to know that since it is most likely a bug or a problem with the dataset. Also, if we check the return code, somebody who wants to make it always pass can do that by making the `startup_command` always return exit code 0, but somebody who wants to be alerted cannot do that if we don't check.

Ideally if the startup command fails, an exception will be added to the list of exceptions that the `swebench.py` script maintains / prints. The way I did it in my code was
```
assert result["returncode"] == 0, result["output"]
```
Probably AssertionError is not the right exception here but directionally this is very useful.

pcmoritz (approved):
Thanks a lot for adding this! I left a comment about surfacing an error if the startup command fails, that would be great to fix before merging it in :)

src/minisweagent/run/extra/swebench.py
klieret: Great point, let's do that!

PR #508, opened 2025-09-08
feat(extra): add Codex CLI config and docs for CLI‑native agents on SWE‑bench

ai-jz:
## Motivation
As stated in the README, mini-swe-agent "makes it perfect as a baseline system and for a system that puts the language model (rather than the agent scaffold) in the middle of our attention."

This PR extends that ethos to support CLI‑native agents (Codex CLI, Claude Code, Gemini CLI,  etc.) by standardizing the harness (swe-bench dataset, containerized environment, and submission outputs) without modifying the CLIs' own internal scaffolds. This enables apples‑to‑apples comparisons between mini-swe-agent's model‑centric baseline and top‑tier CLI‑native agents from OpenAI/Anthropic/Google.

## What's included
- **New config**: `src/minisweagent/config/extra/swebench_codex_cli.yaml` implementing a deterministic two‑step flow (run CLI → submit patch) with prompt setup and optional host mounts for speed.
- **Docs**: A new "CLI‑Native Agents (Codex, Claude Code, Gemini)" section in `docs/usage/swebench.md` with concrete usage and generalization guidance.

## Benefits
- **Apples‑to‑apples harness**: run and compare CLI‑native agents and mini-swe-agent with different models under the same dataset/container/submission conditions.
- **Apples‑to‑apples outputs**: identical `preds.json` → evaluate with `sb-cli` or the SWE‑bench harness.
- **Config‑only integration**: isolated to yaml config in `config/extra/`; no Python code changes to keep mini-swe-agent's core value as it is.

## Scope / Non‑goals
- Does not alter or re‑implement CLI‑native scaffolds; invokes them as‑is in yolo/non‑interactive mode.

## Validation
- [x] Verified deterministic two‑step sequence for Codex CLI
- [x] Created `preds.json` output format
- [x] Documented batch and single‑instance commands
- [x] Tested both host‑mount and container‑install approaches

### Performance:
Codex CLi with `gpt-5-mini` resolves 316 out of 500 instances (63.2% solve ratee) in case of SWE-Bench Verified. As a comparison,  mini-swe-agent with the same model leads to 59.8% solve rate. This minimal performance gap shows the value of mini-swe-bench with its design principals.

## Future work (optional)
- Provide ready‑to‑use `swebench_claude_code_cli.yaml` and `swebench_gemini_cli.yaml` following the same pattern.
- Parse the output of CLI-native code agents into the same format of message history as mini-swe-agent. This will facilitate trajectory-level research in the comparison between mini-swe-agent and CLI-native agent .

### DIFF
--- a/docs/usage/swebench.md
+++ b/docs/usage/swebench.md
@@ -148,8 +148,8 @@ Note: CLI‑native agents include their own internal scaffolds/orchestration. Th
 
 ### Requirements
 - API key set via `mini-extra config set` or exported in the following way:
-  - Codex: `export OPENAI_API_KEY=...` 
-  - Claude Code: `export ANTHROPIC_API_KEY=...` 
+  - Codex: `export OPENAI_API_KEY=...`
+  - Claude Code: `export ANTHROPIC_API_KEY=...`
   - Gemini CLI: `export GEMINI_API_KEY=...`
 - Optional host mounts for faster startup (already encoded in the config): host `node` and `@openai/codex` modules. If not present, the startup script installs them inside the container.
 
--- a/src/minisweagent/config/extra/swebench_codex_cli.yaml
+++ b/src/minisweagent/config/extra/swebench_codex_cli.yaml
@@ -56,9 +56,9 @@ run:
 
     <instructions>
     Please help implement necessary changes to meet requirements in the PR description.
-    Your task is specifically to make changes to non-test files in the current directory 
-    in order to fix the issue described in the PR description in a way that is general 
-    and consistent with the codebase. 
+    Your task is specifically to make changes to non-test files in the current directory
+    in order to fix the issue described in the PR description in a way that is general
+    and consistent with the codebase.
 
     ## Important Boundaries
     - MODIFY: Regular source code files in /testbed (this is the working directory for all your subsequent commands)
@@ -70,8 +70,8 @@ run:
     3. Edit the source code to resolve the issue
     4. Verify your fix works by running your script again
     5. Test edge cases to ensure your fix is robust
-    
-    Before you are finished, clean up the codebase to remove all temporary files. Only keep 
+
+    Before you are finished, clean up the codebase to remove all temporary files. Only keep
     the minimal changes that are ready to be committed, and do not commit yet.
 
     ## Environment Details


klieret:
Hi @ai-jz!

haha, that's an interesting way of using mini! 

However as for actually merging this, I have the following two concerns

- Cost/step limits don't seem to be passed on (?)
- Runtime timeouts aren't passed on 

I can see how this can be a shortcut to evaluate on all of SB, but there seem to be quite some caveats. I think for something like this it might make more sense to link to a blog post of yours or something (?)

ai-jz:
> haha, that's an interesting way of using mini!
> 
> However as for actually merging this, I have the following two concerns
> 
> * Cost/step limits don't seem to be passed on (?)
> * Runtime timeouts aren't passed on
> 
> I can see how this can be a shortcut to evaluate on all of SB, but there seem to be quite some caveats. I think for something like this it might make more sense to link to a blog post of yours or something (?)

@klieret  For step/cost/time limits, you’re right they are important! It's a pity that these aren’t universally supported by these CLI-native agents yet (see details below). mini-swe-agent has default `timeout` of [30 minutes](https://github.com/SWE-agent/mini-swe-agent/blob/639744e5cea94a2bc9a25c12b29317273b7d5eb6/src/minisweagent/environments/docker.py#L22) to run commands -- this serves as a limit to run such agent in a single command in their yolo mode. 

In the short term, is it reasonable to use this conservative timeout (to avoid infinite loops or runaway cost) and, where available, optionally pass through turn/step limits to the underlying CLI agent (Gemini CLI and Claude Code support turns today)? I think we don't need to be blocked on perfect cost controls: timeout-only limit is a form of apple-to-apple among CLI-native agents, and seems fair to models with mini-swe-agent with steps_limit = 250 and cost of $3. Of course, this caveat should be clearly noted in any comparison results. Let me know if you have better suggestions here.

And yes, the blog post idea is great! One idea is to consistently evaluate fast-evolving model progress for coding agents (models via mini-swe-agent) and include reproducible comparisons with top-tier async agents, especially the open-source ones like codex cli and gemini cli. It could also help the community by making related trajectories and differences visible for further study. If you have any suggestions or perspectives on what would be most valuable from the mini-swe-agent point of view, I’d love to discuss more!

------------------

Codex CLI
- Limits: No step/cost limits yet. It's been requested at https://github.com/openai/codex/issues/3201
- Usage: No /cost or /stats yet
Claude Code: 
- Limits: Support step limits (--max-turns), but not cost limit yet
- Usage: Allow to get cost via /cost 
Gemini CLI
- Limits: Support step limits (--max-turns). Not cost limit yet
- Usage: Support /stats that include token usage, but not cost yet (requested at https://github.com/google-gemini/gemini-cli/discussions/4489)

ai-jz:
In fact, it would be meaningful to start including top-tier async agents on the SWE-bench leaderboard, just as you’ve already done by listing leading models in the new leaderboard tab for mini-swe-bench. Many top providers have published impressive SWE-bench scores, but it’s rare to see reproducible code or transparent benchmarks for their agentic pipelines. mini-swe-agent provide a clean method to reproduce the results of these async agent via CLI command.

How do you think establishing this kind of leaderboard - highlighting the best open async/CLI-native solutions and their continuous progress after major updates - could serve as a “north star” for the industry and the developers who are using them more and more often then before? It feels like a natural next step to show not just what they reported, but to provide an objective measure of how they actually performs and advances over time. 

Presenting and minimizing the remaining gap between mini-swe-agent and these top-tier agent is even more meaningful to the community. In fact, although such agent shows better solve rate on swe-bench, it will be interesting to show them on a Pareto frontier of solve rate vs. cost. mini-swe-agent's performance is likely very attractive on such a Pareto frontier. For now, mini-swe-agent with gpt-5-mini-high will be outstanding there 😊

klieret:
Hi @ai-jz thanks again for opening this and all the information. I'm currently closing most old PRs. I think mini-swe-agent should stay as simple as possible, but if you have some blog or fork or something to link to let me know, happy to link to it!

PR #511, opened 2025-09-09
User docker.io as image prefix for swebench images

pcmoritz:
This PR simplifies the code a little bit and also makes sure it works out of the box if `podman` is used instead of `docker`.


klieret (approved):
Thanks, didn't know about this!

PR #513, opened 2025-09-09
Enh(litellm): Include full litellm response

klieret:


src/minisweagent/models/litellm_model.py
aur3l14no, line 1: @klieret maybe we should use `response.json()` here since `ResponseModel` cannot be serialized and `save_traj` would break?

klieret: oops, my bad, let me double check that

PR #517, opened 2025-09-10
Enh: Add a timeout argument as an override for `env.execute`

SumanthRH:
# What does this PR do?

Adds a `timeout` argument for `env.execute` so that some execute commands can use a different timeout setting than the one provided in the config. 

This will support executing arbitrary commands (like running a test) with Mini-SWE-Agent's Environment class

### DIFF
--- a/src/minisweagent/environments/docker.py
+++ b/src/minisweagent/environments/docker.py
@@ -73,7 +73,7 @@ class DockerEnvironment:
         self.logger.info(f"Started container {container_name} with ID {result.stdout.strip()}")
         self.container_id = result.stdout.strip()
 
-    def execute(self, command: str, cwd: str = "", timeout: int | None = None) -> dict[str, Any]:
+    def execute(self, command: str, cwd: str = "", *, timeout: int | None = None) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


src/minisweagent/environments/docker.py
```diff
@@ -89,7 +89,7 @@ def execute(self, command: str, cwd: str = "") -> dict[str, Any]:
         result = subprocess.run(
             cmd,
             text=True,
-            timeout=self.config.timeout,
+            timeout=timeout if timeout is not None else self.config.timeout,
```
pcmoritz: All of these should be written more idiomatically as `timeout or self.config.timeout`

pcmoritz:
Thanks for the review, I updated the PR now.

@klieret Are you ok with me merging PRs that are approved?

klieret:
> @klieret Are you ok with me merging PRs that are approved?

Sounds good, thanks!

klieret:
Only small request would be to update the PR description to start with Feat/Fix/Enh/Doc etc. Makes it easier when compiling the release notes to quickly skim what needs to be included.

Also I prefer to do the squash merge

PR #530, opened 2025-09-16
Handle whitespaces in bash command line

riazarbi:
I have been having a lot of trouble running mini-extra swebench with a llama.cpp-served version of  Qwen3-Coder 30B. Reviewing the trajectories revealed frequent failure to call bash. The reason was that the LLM frequently added whitespace on the bash backtick line, like so:

"```bash  \n"

and this was not caught by the regex, which expects

"```bash\n"

This PR allows for whitespaces on the bash backtick line. I can't see the harm, it just helps along any formatting issues that the LLM sends.


klieret:
Included in v1.12.0, thank you :)

PR #536, opened 2025-09-20
feat: add Requesty model provider support

Thibault00:
- Add RequestyModel class with OpenAI-compatible API support
- Implements cost tracking from API response usage.cost field
- Includes custom headers for GitHub referer and mini-swe-agent title
- Supports all Requesty models via router.requesty.ai/v1 endpoint
- Add requesty to model class mapping for --model-class requesty usage
- Follows same patterns as OpenRouter model implementation


klieret (approved):
Thank you! Sorry for the late approval

PR #551, opened 2025-09-30
Docs: concrete example using a local model

RicardoDominguez:
Using a local model to generate SWE-bench trajectories. Used a 1.7B model as an example such that it is most accessible.


klieret (approved):
Sorry for the late review, thank you & welcome on board :) !

Issue #293, opened 2025-07-25
mini prompt: Ask model to summarize changes at the end


klieret:
needs to be done in a way that doesn't interfere with quitting behavior

PR #556, opened 2025-10-02
Add summarize_on_exit option for mini agent

Chesars:
##Summary

Adds optional summarize_on_exit config to mini agent.
 
When enabled, the model generates a summary of changes before the exit prompt.

  **Changes:**
  - Added `summarize_on_exit` field to `InteractiveAgentConfig` (default:  false)
  - Implemented `_summarize_changes()` method that queries the model for a  summary
  - Summary displays before the exit confirmation prompt
  - Added config example in `mini.yaml` (commented out by default)

  **Usage:**
 Set `summarize_on_exit: true` in any config that uses `InteractiveAgent` (mini.yaml, github_issue.yaml, etc.) 

#293

### DIFF
--- a/src/minisweagent/agents/interactive.py
+++ b/src/minisweagent/agents/interactive.py
@@ -30,6 +30,8 @@ class InteractiveAgentConfig(AgentConfig):
     """Never confirm actions that match these regular expressions."""
     confirm_exit: bool = True
     """If the agent wants to finish, do we ask for confirmation from user?"""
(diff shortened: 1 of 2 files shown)


klieret:
Hi @Chesars Super appreciate this contribution, but for v2 I made the tough decision of removing some things in order to keep mini as slim and maintainable as possible and focus on the core use case. So unfortunately I have to close this. Sorry!

PR #564, opened 2025-10-03
feat: make action parsing regex configurable

Chesars:
## Summary
 Makes the action parsing regex configurable.
  
## Changes

  - Add `action_regex` field to `AgentConfig`
  - Update `parse_action()` to use `self.config.action_regex`
  - Add documentation with XML example

  ## Usage

  ```yaml
  agent:
    action_regex: "<action>(.*?)</action>"
    system_template: |
      Use <action> tags for commands.
      <action>ls -la</action>
  ```
      
No breaking changes.
 
Implements #144

### DIFF
--- a/docs/advanced/yaml_configuration.md
+++ b/docs/advanced/yaml_configuration.md
@@ -32,6 +32,144 @@ You can find the full list of options in the [API reference](../reference/agents
 To use a different agent class, you can set the `agent_class` key to the name of the agent class you want to use
 or even to an import path (to use your own custom agent class even if it is not yet part of the mini-SWE-agent package).
 
+### Custom Action Parsing
+
+By default, mini-SWE-agent parses actions from markdown code blocks (` ```bash...``` `).
+You can customize this behavior by setting the `action_regex` field to support different formats like XML.
+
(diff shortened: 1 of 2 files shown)


klieret (approved):
Thank you for this, especially for also taking care of documentation.

So far we never needed this, but I guess with minimax2 absolutely refusing to do the triple-backtick formatting, this becomes relevant, so let me merge this.

It would also allow us to avoid issues with mini not being able to edit readme files that contain ` ```bash ` by switching to another regex without breaking backwards compatibility

Chesars:
> Thank you for this, especially for also taking care of documentation.
> 
> So far we never needed this, but I guess with minimax2 absolutely refusing to do the triple-backtick formatting, this becomes relevant, so let me merge this.
> 
> It would also allow us to avoid issues with mini not being able to edit readme files that contain ` ```bash ` by switching to another regex without breaking backwards compatibility

Hi Kilian. You’re right, and I didn’t know that Minimax2 require another output format for bash commands and it’s true that it will improve mini to edit files in Markdown like ```bash. Thanks to you!

PR #566, opened 2025-10-08
Fix: Command palette crashes with TypeError in visual mode

Chesars:
## Summary
  Fixed a bug where all commands in the command palette (Ctrl+) would crash with TypeError: the first argument must be callable when selected in visual mode (mini -v).
  
## Details
The get_system_commands() method was passing action names as strings (e.g., "previous_step") to SystemCommand, but Textual's API requires callable methods. When the command palette tried to execute these commands via call_later(), it failed because strings are not callable.

Modified src/minisweagent/agents/interactive_textual.py to use getattr() to convert action strings to their corresponding callable methods (e.g., "previous_step" → self.action_previous_step).

This affects all command palette commands including:
 - Step++ (next step)
  - Step-- (previous step)
  - Step=0 (first step)
  - Step=-1 (last step)
  - Scroll down
  - Scroll up
  - Quit
  - YOLO mode
  - Human mode
  - Confirm mode
  - Show help panel
  - Hide help panel

The fix uses getattr() to convert action strings like "previous_step" to their corresponding callable methods like self.action_previous_step.

Tested in mini -v mode - all command palette commands now execute successfully.

### DIFF
--- a/tests/agents/test_interactive_textual.py
+++ b/tests/agents/test_interactive_textual.py
@@ -863,10 +863,10 @@ async def test_system_commands_are_callable():
     )
 
     async with app.run_test() as pilot:
-        from textual.screen import Screen
-
         screen = app.screen
         commands = list(app.get_system_commands(screen))
 
         for command in commands:
-            assert callable(command.callback), f"Command '{command.title}' has non-callable callback: {command.callback}"
+            assert callable(command.callback), (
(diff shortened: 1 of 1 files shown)


klieret:
Oh wow, I wasn't aware of this! Thank you so much for not only observing this, but also following up with the fix, really appreciated! Let me merge & release

Chesars:
You're welcome Kilian 🙌.

PR #567, opened 2025-10-08
Enh: Allow extending cost limit in textual UI

### LINKED ISSUE #335: Enh: Extend cost limit in textual

Chesars:
## Summary
 Currently, in the textual UI (`mini -v`), the agent stops immediately when the cost/step limit is exceeded  (default $3.00). This PR adds the ability for users to extend these limits on-the-fly when using `mini -v` (textual UI), allowing tasks to continue without restarting.

## Details

  - Added `LimitsExceeded` exception handling in `TextualAgent.query()`
  - When limits are exceeded, shows a warning notification with current limits and spend
  - Prompts user for new step limit and cost limit with helpful placeholders
  - Added optional `placeholder` parameter to `SmartInputContainer.request_input()`

## Testing

  - Updated `test_agent_with_cost_limit` and `test_agent_with_step_limit`
  - All 25 tests in `test_interactive_textual.py` pass

Closes #335

### DIFF
--- a/tests/agents/test_interactive_textual.py
+++ b/tests/agents/test_interactive_textual.py
@@ -407,13 +407,16 @@ async def test_agent_with_cost_limit(default_config):
         assert app.agent_state == "RUNNING" or app.agent_state == "AWAITING_INPUT"
 
 
-async def test_agent_with_cost_limit_invalid_input():
+async def test_agent_with_cost_limit_invalid_input(default_config):
     """Test ValueError handling when invalid input is provided for limits - now retries with while loop."""
     app = TextualAgent(
         model=DeterministicModel(outputs=["Response 1", "Response 2", "Response 3"]),
         env=LocalEnvironment(),
-        mode="yolo",
-        cost_limit=0.01,
+        **{
+            **default_config,
+            "mode": "yolo",
+            "cost_limit": 0.01,
+        },
     )
 
     app.notify = Mock()


src/minisweagent/agents/interactive_textual.py
```diff
@@ -57,7 +57,33 @@ def query(self) -> dict:
             self.add_message("assistant", msg["content"])
             return msg
         self._current_action_from_human = False
-        return super().query()
+        try:
+            return super().query()
+        except LimitsExceeded:
+            # Show current limits and prompt for new ones (matching interactive.py behavior)
+            # Show notification with the limits info
+            self.app.notify(
```
klieret: Can we move that into the prompt, i.e., the first prompt is

"Some limit was exceeded. Current number of steps xx, current limit yy. Enter new limit (0 for unconstrained)"

and I expect that the step limit is probably always 0, so it might be good to test if "step limit == 0" (we skip asking for a new step limit, because that's not the one we blew).

src/minisweagent/agents/interactive_textual.py
```diff
@@ -57,7 +57,33 @@ def query(self) -> dict:
...
+                severity="warning",
+                timeout=10,
+            )
+
+            # First prompt: new step limit (short message near input)
+            new_step_limit = self.app.input_container.request_input("New step limit:", placeholder="0 for unlimited")
+            try:
+                self.config.step_limit = int(new_step_limit) if new_step_limit.strip() else 0
+            except ValueError:
+                self.config.step_limit = 0
```
klieret: this is dangerous, we can't do that, users would think they have set a limit but they haven't, honestly I'd rather have the application crash.

Or we put it in a small while loop that only breaks once it works.

Chesars: You're right, let's put a while loop.

src/minisweagent/agents/interactive_textual.py
Chesars: Okay, great, let's move that to the prompt, leave the placeholder as it was, and remove the notification for simplicity. Yes, it makes sense that the step limit probably is always 0, so let's add an if statement for that.

tests/agents/test_interactive_textual.py
```diff
@@ -343,39 +343,105 @@ async def test_agent_with_cost_limit():
...
+        # Simulate user entering new limits
+        await pilot.press("1", "0")  # step limit = 10
+        await pilot.press("enter")
+        await pilot.pause(0.2)
+
+        # Now it should ask for cost limit
+        assert app.input_container.pending_prompt == "New cost limit:"
+
+        await pilot.press("1", "0")  # cost limit = 10
+        await pilot.press("enter")
```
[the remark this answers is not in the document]
Chesars: Applied 👍

Chesars:
> This is great, I just tested it and it seems to work very well! I left two things that need to be resolved though, but shouldn't be too hard

Thanks Kilian 🙌. Great, I applied the suggestions there, they make sense, sorry for some oversights, I tried it and it seems to be working great 👍.

src/minisweagent/agents/interactive_textual.py
```diff
@@ -57,7 +57,36 @@ def query(self) -> dict:
...
+            # Show current limits and prompt for new ones (matching interactive.py behavior)
+            # First prompt: new step limit (only if there is a step limit configured)
+            if self.config.step_limit > 0:
+                while True:
+                    new_step_limit = self.app.input_container.request_input(
+                        f"Step limit exceeded. Current: {self.model.n_calls} steps, "
+                        f"limit: {self.config.step_limit}. Enter new limit (0 for unlimited):"
+                    )
+                    try:
+                        self.config.step_limit = int(new_step_limit) if new_step_limit.strip() else 0
```
Chesars, line 70: Quick question about the limit behavior here (applies to both step_limit and cost_limit):

 Currently, we're setting the new limit as an absolute value. So if someone has already used 3 steps with a limit of 3, and they enter "3" again, the agent stops
  immediately since 3 >= 3.

  This might be confusing for users. A couple options we could consider:

  1. **Show a warning and let them retry** - if `new_limit <= current_usage`:
     ```python
     self.app.notify(
         f"Warning: New limit ({new_limit}) must be greater than current usage ({self.model.n_calls}). Please try again.",
         severity="warning"
     )
     continue  # Stay in the while loop, ask again

  2. Make it incremental - so when they type "3", it means "3 MORE steps":
    - Update the prompt to say something like "How many additional steps?"
    - Then do: self.config.step_limit = self.model.n_calls + int(new_step_limit)

 WDYT? Should we implement one of these improvements?

klieret:
Thanks for the updates, will test this soon!

Chesars:
> Thanks for the updates, will test this soon!

Ok, Great 🙌.

klieret:
Hi @Chesars . thanks again for all of the work. In the end, I decided v2 won't have the visual interface anymore. I just feel like it wasn't really used (because it's also very different from most other agent CLIs like claude code etc.) and was disproportionately difficult to maintain. Sorry, I know you put some time on it.

Chesars:
Hi @klieret. Makes sense - that makes mini-swe-agent v2 simpler and with better maintainability. Thanks for being so welcoming - it's been a pleasure contributing here.

PR #568, opened 2025-10-09
feat: enable interrupt for long-running commands in textual UI (mini -v)

### LINKED ISSUE #332: Enh: Allow to interrupt agent run in textual mode

Chesars:
## Summary 
Implements immediate interrupt functionality for interactive textual mode, matching the behavior of interactive.py. When the user presses 'i', the agent now interrupts immediately, even during long-running commands.

## Changes:
- Modified execute_action() to use subprocess.Popen() with polling instead of env.execute(), enabling process termination on interrupt
- Added _current_process tracking to _TextualAgent
- Modified action_interrupt() to terminate running subprocess before injecting KeyboardInterrupt
- Added KeyboardInterrupt handlers in query() and execute_action()
- Added tests for interrupt during command execution

The interrupt now works for:
- Long-running commands.
- LLM queries.
- Any point in the agent loop.

Closes #332

### DIFF
--- a/docs/usage/mini_v.md
+++ b/docs/usage/mini_v.md
@@ -49,6 +49,7 @@ Useful switches:
 
 - `f1` or `?`: Show keybinding help
 - `q` (or `ctrl+q`): Quit the agent
+- `i` (or `ctrl+i`): Interrupt the agent immediately (works during long-running commands)
 - `c`: Switch to `confirm` mode
 - `y` (or `ctrl+y`): Switch to `yolo` mode
 - `h` or `LEFT`: Go to previous step of the agent


src/minisweagent/agents/interactive_textual.py
```diff
@@ -83,7 +110,74 @@ def execute_action(self, action: dict) -> dict:
             result = self.app.input_container.request_input("Press ENTER to confirm or provide rejection reason")
             if result:  # Non-empty string means rejection
                 raise NonTerminatingException(f"Command not executed: {result}")
-        return super().execute_action(action)
+
```
klieret, line 113: no, we can't do this like that,this completely bypasses the environment setup (we might not be running locally but in a docker container for example). I thought it might be possible to just catch the keyboardinterrupt at the same place? Take a look at how this is done for the non-visual interactive mode

Chesars: You're right. This bypasses the environment setup. We can catch the keyboardinterrupt but Textual (the Python library used) sets the terminal in Raw Mode, so does not terminate the subprocess when Ctrl + C is pressed. In `interactive.py` (Cooked Mode), the OS handles this at the kernel level. In environment files the subprocess is executed using `.run()` ( a function from Python's subprocess module which internally uses `.wait()`), the subprocess blocks at the kernel level and we cannot interrupt it from Python. A solution could be to change to `subprocess.Popen()` (Process Open) which returns a process object, allowing us to use`.terminate()`, `.wait()`, `.poll()`, and other methods (which are Python wrappers around kernel syscalls). We could implement this change in the environment files themselves, each environment would need to expose a execution method using `Popen`. The current implementation works correctly for LocalEnvironment but bypasses the environment abstraction for others. I'm happy to collaborate on implementing this properly.

klieret:
Copying this from the other PR, but again super sorry that I have to close this

> Hi @Chesars . thanks again for all of the work. In the end, I decided v2 won't have the visual interface anymore. I just feel like it wasn't really used (because it's also very different from most other agent CLIs like claude code etc.) and was disproportionately difficult to maintain. Sorry, I know you put some time on it.

PR #591, opened 2025-11-14
Fix: Allow free OpenRouter models with zero cost

### LINKED ISSUE #583: Free Openrouter models models do not work with mini-swe-agent
### Describe the bug

I tried running mini-swe-agent with minimax2 to benchmark it. I tried using a dummy cost in model_kwargs to bypass it and also tried `export MSWEA_DISABLE_COST_TRACKING=true`

Image

Chesars:
## Summary
Fixes #583 - Allows free OpenRouter models (like minimax2) to work by removing overly restrictive zero-cost validation.

## Details
Free models on OpenRouter legitimately return `cost = 0.0` according to their usage tracking feature. The existing code rejected these models with an error, preventing users from using free-tier models.

## Solution
- Removed the `if cost == 0.0` check that raised `OpenRouterAPIError`
- Kept the `assert cost >= 0.0` validation to prevent negative costs
- Now consistent with `LitellmModel` and `PortkeyModel` implementations

### DIFF
--- a/tests/models/test_openrouter_model.py
+++ b/tests/models/test_openrouter_model.py
@@ -7,7 +7,6 @@ import requests
 
 from minisweagent.models import GLOBAL_MODEL_STATS
 from minisweagent.models.openrouter_model import (
-    OpenRouterAPIError,
     OpenRouterAuthenticationError,
     OpenRouterModel,
 )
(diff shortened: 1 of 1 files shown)


klieret:
Hi @Chesars , I appreciate this, but I just think that ensuring we don't mess up cost tracking is super crucial (and in the current implementation, if we don't get any cost field, it's also 0., which motivated the `==0` check). I'll implement an env variable to unset the strict behavior, but otherwise leave it by default. 

I'll start with this commit, so you still get authorship on the commit and I can adapt your new test.

klieret:
Superseded by #593

Chesars:
> Hi @Chesars , I appreciate this, but I just think that ensuring we don't mess up cost tracking is super crucial (and in the current implementation, if we don't get any cost field, it's also 0., which motivated the `==0` check). I'll implement an env variable to unset the strict behavior, but otherwise leave it by default. 
> 
> 
> 
> I'll start with this commit, so you still get authorship on the commit and I can adapt your new test.

Oh that's great. Sure, you 're right that if we don't get any cost field, it's also 0. An env variable would work great for that. Thanks to you!🙌

Chesars:
> Hi @Chesars , I appreciate this, but I just think that ensuring we don't mess up cost tracking is super crucial (and in the current implementation, if we don't get any cost field, it's also 0., which motivated the `==0` check). I'll implement an env variable to unset the strict behavior, but otherwise leave it by default. 
> 
> 
> 
> I'll start with this commit, so you still get authorship on the commit and I can adapt your new test.

Yeah, of course. Great🙌

PR #622, opened 2025-11-25
chore: remove temperature defaults from .yaml files (expect swe-bench.yaml)

Chesars:
## Summary
  - Remove hardcoded `temperature: 0.0` from config files to let models use their defaults
  - Delete unused `mini_no_temp.yaml` config file

  ## Changes
  - `src/minisweagent/config/mini.yaml`: Remove temperature default
  - `src/minisweagent/config/default.yaml`: Remove temperature default
  - `src/minisweagent/config/github_issue.yaml`: Remove temperature default
  - `src/minisweagent/config/mini_no_temp.yaml`: Deleted (redundant)

### DIFF
--- a/docs/models/troubleshooting.md
+++ b/docs/models/troubleshooting.md
@@ -67,7 +67,7 @@ Another common mistake is to not include any or the correct provider in the mode
 
 ## Temperature not supported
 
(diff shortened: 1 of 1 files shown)


klieret:
thx. I think the no temp config file is still referenced in a couple of files. should remove

Chesars:
> thx. I think the no temp config file is still referenced in a couple of files. should remove

You're right, there I updated the references. You're welcome.

PR #636, opened 2025-12-13
Update swebench submission instructions to add specific files

aorwall:
Changed the submission prompt to instruct the agent to add specific modified source files instead of using `git add -A`, which could include unrelated files like compiled code, cache files, or other artifacts.


klieret:
OK! I'd say we hold this for now until we finalized the SB ML prompt and then do a consistency check that this doesn't change the SB V results too much (just being super conservative here)

klieret:
(merged in v2.0)

PR #637, opened 2025-12-13
Fix: Remove timestamp field from OpenAI Response API requests

aorwall:
## Summary
- The `timestamp` field added to messages by the agent is not supported by the OpenAI Response API, causing API errors
- Removed the `timestamp` field from messages before sending requests to the Response API
- Also cleaned up a debug `print()` statement


klieret (approved):
thanks, good catch!

klieret (approved):

PR #638, opened 2025-12-14
Reraise original exception when retries are exhausted

### LINKED ISSUE #635: Exceptions in LM query almost always raised as RetryError
It seems right now that almost all errors (internal server exception, context lim exceeded etc.), might be raised as retryerror in the end, which is bad for bookkeeping

aorwall:
Add reraise=True to all @retry decorators so that when retries are exhausted, the original exception is re-raised instead of being wrapped in a RetryError. 

Fixes #635


klieret:
nice! exactly the kind of simple solution I hoped for

Issue #659, opened 2025-12-19
Change exit sequence order (command && echo MINI_SWE_AGENT_FINAL_OUTPUT)


klieret:
Currently we do `echo MINI_SWE_AGENT_FINAL_OUTPUT && command for final output`.

However, this is bad, because if the command fails (e.g., an incorrect git commit), we still exit the agent.

Options to solve:

- Change order: `command && echo ...` (this is also a backwards compatibility breaking change, though it's also very obvious for users if they forgot to update their prompts for this, because they'll get 0%)
- Explicitly check exit status

PR #661, opened 2025-12-19
fix: check exit code before accepting sentinel in has_finished()

Chesars:
## Summary
- Check returncode before accepting the sentinel in `has_finished()`
- If the command fails (non-zero exit code), the agent continues instead of terminating, allowing it to see the error and retry

## Details
Currently, if a command like `echo COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT && git add -A && git diff` fails after the echo, the agent still terminates because it already saw the sentinel. 

With this fix, we check the exit code first. If it's non-zero, we ignore the sentinel and let the agent continue, so it can see the error and attempt to recover.

This implements "Option 2" from #659 (explicitly check exit status).

## Note on breaking changes
Unlike Option 1 (changing command order), this approach is **not a breaking change** - existing prompts continue to work as before.

## Test plan
- [x] Added `test_does_not_finish_on_failed_command` test
- [x] All existing tests pass

### DIFF
--- a/tests/agents/test_default.py
+++ b/tests/agents/test_default.py
@@ -315,7 +315,7 @@ def test_step_output_includes_action(default_config):
     assert "output" in output
 
 
-def test_does_not_finish_on_failed_command():
+def test_does_not_finish_on_failed_command(default_config):
     """Test agent continues when command with sentinel fails (non-zero exit code)."""
     agent = DefaultAgent(
         model=DeterministicModel(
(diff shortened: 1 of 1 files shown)


klieret:
hi @Chesars thanks for this! 

Right now I've opened a bunch of issues that I have assigned to myself because for mini v2.0 there's a lot of decisions to make. So for these issues, still a lot of things can change and they might be larger refactors. So more unlikely that I will merge PRs for them, because I'm working on them already/there's still some decisions to take. you can check if the issues are already assigned to someone, it shows on the right

Chesars:
Hi @klieret . Got it, I'll check for assignments before picking up issues. Thanks for the clarification.

Chesars:
Feel free to close, review, refactor, or whatever with this PR, no problem. Thanks!

Issue #554, opened 2025-10-01
Integrating mini-swe-agent with Modal/swe-rex


mohit-raghavendra:
### Describe the issue

Hi, is it possible to have mini-swe-agent integrated with a modal runtime, perhaps using swe-rex? SWE-Agent already has this, and getting remote execution can help in training coding agents, similar to the setup in the CWM paper

klieret:
Hi yes, there's already one class in `environments/extra` though it's just for local docker, I just haven't done it. But it's pretty trivial, you basically just need a small wrapper (you can see how it's done with the existing swe-rex implementation).

klieret:
I'm hoping to get to this maybe by next week

PR #679, opened 2026-01-04
Integrate Modal for scalable remote execution

AlienKevin:
This PR implements https://github.com/SWE-agent/mini-swe-agent/issues/554 and supports running mini-swe-agent on Modal via SWE-Rex.

Tested on SWE-bench with GPT-5 mini:
```
mini-extra swebench \
    --config src/minisweagent/config/extra/swebench_modal.yaml \         
    --subset verified \
    --split test \
    --workers 100 \
    -o ./results/gpt5-mini-modal
```

[swe-bench_verified__test__gpt5-mini-modal.json](https://github.com/user-attachments/files/24419971/swe-bench_verified__test__gpt5-mini-modal.json)
[trajectories.zip](https://github.com/user-attachments/files/24419972/results.zip)

### DIFF
--- a/src/minisweagent/run/extra/swebench.py
+++ b/src/minisweagent/run/extra/swebench.py
@@ -81,12 +81,10 @@ def get_sb_environment(config: dict, instance: dict) -> Environment:
     env_config = config.setdefault("environment", {})
     env_config["environment_class"] = env_config.get("environment_class", "docker")
     image_name = get_swebench_docker_image_name(instance)
-    if env_config["environment_class"] == "docker":
+    if env_config["environment_class"] in ["docker", "swerex_modal"]:
         env_config["image"] = image_name
     elif env_config["environment_class"] == "singularity":
         env_config["image"] = "docker://" + image_name
-    elif env_config["environment_class"] == "swerex_modal":
-        env_config["image"] = image_name
     env = get_environment(env_config)
     if startup_command := config.get("run", {}).get("env_startup_command"):
         startup_command = Template(startup_command, undefined=StrictUndefined).render(**instance)


klieret:
Very cool! Thanks for testing this, glad it works. Happy to merge this (or do you want to do more changes?)

Couple of things:
- I might move it to the `environments/extra` dir, just because it's probably still a bit more experimental than the others
- There's an upcoming v2.0 where a couple of things change, but I think it should be trivial to make these changes

src/minisweagent/environments/extra/swerex_modal.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+        return {
+            "output": output.stdout,
+            "returncode": output.exit_code,
+        }
+
+    def get_template_vars(self) -> dict[str, Any]:
+        return self.config.model_dump()
+
+    def stop(self):
```
klieret, line 83: I think we should call this `__cleanup__` and then put it in `__del__` just to stay consistent with e.g., `docker.py`.

src/minisweagent/run/extra/swebench.py
```diff
@@ -85,6 +85,8 @@ def get_sb_environment(config: dict, instance: dict) -> Environment:
         env_config["image"] = image_name
     elif env_config["environment_class"] == "singularity":
         env_config["image"] = "docker://" + image_name
+    elif env_config["environment_class"] == "swerex_modal":
```
klieret: could just combine with line 84: `if .. in `["docker", "swerex_modal"]`

src/minisweagent/run/extra/swebench.py
```diff
@@ -155,6 +158,8 @@ def process_instance(
         exit_status, result = type(e).__name__, str(e)
         extra_info = {"traceback": traceback.format_exc()}
     finally:
+        if env and hasattr(env, "stop"):
```
klieret, line 159: don't think we need this if we just add the stop code to `__del__` as mentioned above

pyproject.toml
```diff
@@ -50,6 +50,8 @@ dependencies = [
 full = [
```
klieret, line 50: best practice would be to add a new `modal = ["modal", "boto3"]` and then add `mini-swe-agent[modal]` here?  (similar to how we do it with `dev`) and then we can tell people in the docs to install `mini-swe-agent[modal]`

klieret (approved):
I added a bunch of nitpicks, lmk if you wanna do them else I can do them as well.

Either way, this lgtm!

src/minisweagent/environments/extra/swerex_modal.py
AlienKevin: I tried calling stop in `__del__` but it wasn't triggered for some reason, so I changed to trigger cleanup manually.

src/minisweagent/run/extra/swebench.py
AlienKevin: I tried calling stop in `__del__` but it wasn't triggered for some reason, so I changed to trigger cleanup manually.

src/minisweagent/run/extra/swebench.py
AlienKevin: Done.

pyproject.toml
AlienKevin: Done.

AlienKevin:
@klieret Thanks for the detailed review! Made a few changes according to your comments.

PR #691, opened 2026-01-07
Avoid submitting on failed command output; return error to model

aorwall:
Today, a failed git command can emit output that gets treated as the final submission, causing us to submit errors instead of patches.

This lets the model receive the error feedback and retry instead of submitting an error string in place of a patch.


klieret:
This is now part of the v1 branch which will be the future branch for important fix releases to v1 (I don't expect many)

klieret:
it's not needed in v2 because we just flipped the order of the git command

klieret:
(I cherry picked it to v1, because I couldn't rebase)

PR #702, opened 2026-01-15
Fix uninitialized error_msg variable in tool call parsing

aorwall:
Initialize error_msg before use to prevent NameError when tool call arguments parse successfully but tool name is invalid.


klieret:
nice catch! test failures should be fixed on v2 branch

PR #705, opened 2026-01-19
chore: add git diff instructions to SWE-bench configs

aorwall:
Add explicit git diff command format with file paths to patch creation instructions

### DIFF
--- a/src/minisweagent/config/extra/swebench.yaml
+++ b/src/minisweagent/config/extra/swebench.yaml
@@ -141,7 +141,7 @@ agent:
 
     Step 1: Create the patch file
     Run `git diff -- path/to/file1 path/to/file2 > patch.txt` listing only the source files you modified.
-    Do NOT commit your changes. NEVER use `diff -u` or backup files.
+    Do NOT commit your changes.
 
     <IMPORTANT>
     The patch must only contain changes to the specific source files you modified to fix the issue.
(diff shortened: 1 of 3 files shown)


src/minisweagent/config/extra/swebench.yaml
```diff
@@ -140,7 +140,8 @@ agent:
     Follow these steps IN ORDER, with SEPARATE commands:
 
     Step 1: Create the patch file
-    Save a git diff of your changes to patch.txt.
+    Run `git diff -- path/to/file1 path/to/file2 > patch.txt` listing only the source files you modified.
+    Do NOT commit your changes. NEVER use `diff -u` or backup files.
```
klieret: I was confused there for a second because I thought this was about `git diff -u` (and apparently the `-u` flag has no effect).

If we recommend this, maybe we can be a bit clearer:

```
Run `git diff -- path/to/file1 path/to/file2 > patch.txt` listing only the source files that you want to include (see below).
This assumes that you have not committed any of your changes before.
Do NOT commit your changes.
NEVER use `diff -u` on backup files instead of `git diff`.
```

src/minisweagent/config/extra/swebench_xml.yaml
```diff
@@ -126,7 +126,8 @@ agent:
     Follow these steps IN ORDER, with SEPARATE commands:
 
     Step 1: Create the patch file
-    Save a git diff of your changes to patch.txt.
+    Run `git diff -- path/to/file1 path/to/file2 > patch.txt` listing only the source files you modified.
+    Do NOT commit your changes. NEVER use `diff -u` or backup files.
```
klieret: let's maybe also add a 'Do NOT commit your changes.' to the critical requirements list above.

src/minisweagent/config/extra/swebench.yaml
aorwall: No, it's just overfitting. I'll remove "NEVER use `diff -u` or backup files." but keep the rest.

PR #709, opened 2026-01-21
fix: handle missing command arg in tool calls

aorwall:
Add validation for the 'command' key in tool call arguments before accessing it. Previously, if the model called the bash tool with arguments missing the 'command' key, it would crash with a KeyError instead of raising a FormatError that could be handled gracefully.


klieret:
ah I guess we might also need to raise a `FormatError` if there is no tool call

PR #719, opened 2026-01-29
Fix: Retry on transient API errors (500, etc.)

aorwall:
Remove APIError from abort_exceptions to enable retries on transient server errors like 500 Internal Server Error. Previously these errors would abort immediately instead of being retried with exponential backoff.


klieret:
oh shoot, I didn't realize this was now aborting, I remember this being transient on some models, too

PR #720, opened 2026-01-30
Fix: Update model-class option flag in swebench scripts to avoid conflict

bofia:
In the swebench extras: `config_path` arg and `model_class` both use the `-c` shorthand, creating a conflict


klieret:
Wow, thanks for spotting this, never noticed that! 
Didn't affect the v2 branch which will be fully released on Monday, but I'll make sure to also release a v1 fixup for this

klieret:
This is now part of v1.17.5

thx again and welcome on board :)

PR #721, opened 2026-01-30
feat: Add LLM response streaming to avoid HTTP read timeouts

SailorJoe6:
When using locally-hosted vLLM models, long generations can cause the HTTP connection to timeout before a response is received, triggering unnecessary retries. Streaming mode receives tokens as they're generated, keeping the connection alive and avoiding the timeout retry loop.  This is particularly bad when trying to use mini-swe-agent to troubleshoot a local LLM setup, as they can begin to hallucinate causing an almost eternal timeout/retry loop.  

- Add use_streaming config option (default: true, env: MSWEA_USE_STREAMING)
- Implement _reconstruct_response_from_stream() to accumulate streaming chunks
- Prevents HTTP read timeouts on long vLLM generations (>10 minutes)
- Keeps connection alive by receiving tokens incrementally

### DIFF
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import re
+import time
 from collections import Counter
 from collections.abc import Callable
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


klieret:
Hmm, I'm surprised that there's these timeouts. Like especially for local models where you shouldn't lose connection. I don't think we had timeouts when calling cloud based services. Is there maybe some way to handle this from vllm side?
I'm hesitant to merge this because it makes the class a lot more complicated, especially for what we treat as the default class.

SailorJoe6:
The timeouts occurred for me when testing Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 hosted by vLLM on my Nvidia DGX Spark.  Maybe it's a bit of an edge case, but a real issue.  The reason for the timeout was that the model would often hallucinate and send a never-ending sequence of closing XML tags at the end of the response.   Since this would cause a timeout / retry loop, the test ran for hours with no visible outputs to help diagnose what was going wrong.  I had to add this patch to stream in order to stop the timeout/retry loop and get the full response before I could see that it was hallucinating all this XML.  I also have a patch stacked on this one that streams the trajectory out to disk, rather than holding it in memory and dumping it at the end of the run.  that was also necessary in order to tail the trajectory logs and see what the agent was doing.

SailorJoe6:
All that being said, I've had to jump through some hoops after syncing my fork, and now the commit seems a bit... messy?  If you prefer I could re-implement this on the latest main and try to make it simpler, more targeted.  Or, if you would rather not address this issue, I could rebase the streaming trajectory logic on main and submit that as a patch (assuming you want that).  

I have to say though, troubleshooting the issue of a runaway hallucinating LLM without these two changes was just impossible.

SailorJoe6:
Nevermind.  I think I need to close this, as something weird crept in when I tried to sync my fork.  Back to the drawing board.

PR #726, opened 2026-02-04
Fix swebench-single output path

### LINKED ISSUE #725: swebench-single ignores the -o/--output flag.
### Describe the bug

  swebench-single ignores the -o/--output flag. Even when a custom output path is provided, the
  trajectory is not saved to that path. The script builds the agent config without setting
  agent.output_path and never explicitly saves the final trajectory to the requested output path, so the
  output file is missing or only written to the default location.

  Expected behavior:
  Passing -o /path/to/traj.json should save the final trajectory to that exact file.

  Actual behavior:
  No file is written at the requested output path 

  Minimal reproduction:

  mini-extra swebench-single --subset _test --split test -i

SailorJoe6:
Fixes #725 

  # Summary
  swebench-single currently ignores the -o/--output flag and does not write the
  trajectory to the requested path. This PR sets agent.output_path from the CLI --output
  option so the final trajectory is saved where the user specifies.

  # Changes

  - Set agent.output_path from --output in swebench_single.py.
  - Tests: verify the output file is created when --output is passed.

  # Why
  The CLI advertises -o/--output as the destination for the trajectory, but the file is
  not written there. This is confusing for users and breaks scripting workflows.

  # Testing
  python -m pytest -q tests/run/test_swebench_single.py

### DIFF
--- a/src/minisweagent/run/benchmarks/swebench_single.py
+++ b/src/minisweagent/run/benchmarks/swebench_single.py
@@ -74,6 +74,7 @@ def main(
         configs.append({"model": {"model_name": model_name}})
     if exit_immediately:
         configs.append({"agent": {"confirm_exit": False}})
+    configs.append({"agent": {"output_path": output}})
     config = recursive_merge(*configs)
 
     env = get_sb_environment(config, instance)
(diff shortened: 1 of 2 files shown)


klieret:
Thanks, appreciated!

klieret:
congrats to your first commit in mini :) welcome on board

PR #728, opened 2026-02-04
Make trajectory serialization JSON-safe

### LINKED ISSUE #727: Invalid tool call FormatError sometimes breaks trajectory serialization
### Describe the bug

  **Summary**
  When a tool‑calling model returns an invalid tool call, the framework raises a FormatError. The exception payload can include non‑JSON‑serializable
  objects (e.g., raw response objects), which causes agent.save() to fail. As a result, the final .traj.json is never written, even though
  trajectories are normally always persisted.

  **Observed behavior**

  - Invalid tool call → FormatError
  - agent.save() attempts to json.dumps() the trajectory
  - Serialization fails due to non‑JSON‑serializable data in the error payload
  - Trajectory file is not written to disk

  **Expected behavior**

  - Trajectories should always be written, even when the model response is malformed
  - Error payloads should be JSON‑safe

  **Impact**
  Losing the trajectory on malformed tool calls makes debugging and evaluation difficult, and breaks expected SWE‑bench behavior.

  Fix (coming soon in a PR)

  - Wrap FormatError with a JSON‑safe payload (raw assistant content/tool_calls only + minimal metadata)
  - Add a to_jsonable helper and use it before json.dumps() in DefaultAgent.save() so serialization is robust

SailorJoe6:
Fixes #727 

# Summary
When a tool‑calling model emits an invalid tool call, the resulting FormatError can include non‑JSON‑serializable data, which causes agent.save() to fail and prevents the trajectory file from being written. This PR makes the error payload JSON‑safe and hardens serialization so trajectories are always persisted.

# Changes

- Wrap FormatError in LitellmModel with a JSON‑safe debug payload that preserves raw assistant content/tool_calls and response metadata.
- Add to_jsonable utility and use it in DefaultAgent.save() to guarantee JSON serialization of trajectories.

# Why
Losing the trajectory on invalid tool calls breaks expected SWE‑bench behavior and makes debugging difficult. This ensures trajectories are written even on malformed tool calls.

### DIFF
--- a/src/minisweagent/agents/default.py
+++ b/src/minisweagent/agents/default.py
@@ -12,7 +12,7 @@ from pydantic import BaseModel
 
 from minisweagent import Environment, Model, __version__
 from minisweagent.exceptions import InterruptAgentFlow, LimitsExceeded
-from minisweagent.utils.serialize import recursive_merge
+from minisweagent.utils.serialize import recursive_merge, to_jsonable
 
 
 class AgentConfig(BaseModel):
@@ -151,5 +151,5 @@ class DefaultAgent:
         data = self.serialize(*extra_dicts)
         if path:
             path.parent.mkdir(parents=True, exist_ok=True)
-            path.write_text(json.dumps(data, indent=2))
+            path.write_text(json.dumps(to_jsonable(data), indent=2))
         return data
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -9,13 +9,13 @@ from typing import Any, Literal
 import litellm
 from pydantic import BaseModel
 
+from minisweagent.exceptions import FormatError
 from minisweagent.models import GLOBAL_MODEL_STATS
 from minisweagent.models.utils.actions_toolcall import (
     BASH_TOOL,
     format_toolcall_observation_messages,
     parse_toolcall_actions,
 )
-from minisweagent.exceptions import FormatError
 from minisweagent.models.utils.anthropic_utils import _reorder_anthropic_thinking_blocks
 from minisweagent.models.utils.cache_control import set_cache_control
 from minisweagent.models.utils.openai_multimodal import expand_multimodal_content
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -15,6 +15,7 @@ from minisweagent.models.utils.actions_toolcall import (
     format_toolcall_observation_messages,
     parse_toolcall_actions,
 )
+from minisweagent.exceptions import FormatError
 from minisweagent.models.utils.anthropic_utils import _reorder_anthropic_thinking_blocks
 from minisweagent.models.utils.cache_control import set_cache_control
 from minisweagent.models.utils.openai_multimodal import expand_multimodal_content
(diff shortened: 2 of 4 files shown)


klieret:
Thanks for raising this. I'm surprised that FormatError would include any non-serializable content. It should be a very simple text message that doesn't take direct model input. Could it be that adding the full response to the data is what caused this?

klieret:
I feel like for this, adding a new kwarg `response=response.model_dump()` to `parse_toolcall_actions` and then adding that to the extra field when we raise the `FormatError` would be more precise.

klieret:
If there's any other serialization issues, I'd be super interested in squashing them. But I don't think the blanket "try to coerce to json" is maybe the best way. It would point more towards other bugs

SailorJoe6:
I've been using mini-swe-agent to test OSS models hosted locally on a DGX Spark via vLLM, running them against SWE-Bench Multilingual as a first set of tests.  I've also been trying a few different configs, adjusting the system prompt, etc.  Sometimes the LLM hallucinates pretty badly leading to malformed json as the tool call response, this leads to a hard exception and the FormatError actually becomes a textual representation of the exception name in the trajectory, rather than a safely JSON serialized depiction of what the model tried to do.  This code was meant to ensure that no matter what the model tries to do, it's serialized to JSON safely.  

TL/DR - it was the most robust fix I could think of that actually allows you to see what the model spit out.

SailorJoe6:
I do agree that a targeted fix in the tool call parsing path would be better.  I'm happy to update this to attach response.model_dump() on FormatError, rather than the raw response object.  Is that what you had in mind?  

That said, do you think it’s unreasonable to make save() robust against serialization failures?  Trajectories are logging artifacts, and we want them to survive even when the model misbehaves.  How would you feel about keeping a minimal, safe serialization guard in the error wrapper to ensure we get logs?

SailorJoe6:
One more comment for you to consider when you decide how we should handle this.  The model sometimes returns more than one tool call.   I've often seen it return 3 tool calls in one response.  If any one of these has malformed JSON, it still throws the error that causes the framework to ignore the good tool calls and only return the ToolCallError response to the LLM.  

My patch does not address that.  Probably a separate issue.

klieret:
Thanks for all the input!

Let me get back to you for the other comments, soon, but regarding this:

> One more comment for you to consider when you decide how we should handle this. The model sometimes returns more than one tool call. I've often seen it return 3 tool calls in one response. If any one of these has malformed JSON, it still throws the error that causes the framework to ignore the good tool calls and only return the ToolCallError response to the LLM.

I think this is the best we can do. If the model issues toolcalls A B C and A is already broken format, I think it's better to just ask it to reconsider the whole thing. Else, if B and C were depending on A to run and we skip A but then execute B and C, I think this would be even more problematic (theoretically it should only issue separate toolcalls for independent things, but that's also not guaranteed).

SailorJoe6:
> I think this is the best we can do. If the model issues toolcalls A B C and A is already broken format, I think it's better to just ask it to reconsider the whole thing. Else, if B and C were depending on A to run and we skip A but then execute B and C, I think this would be even more problematic (theoretically it should only issue separate toolcalls for independent things, but that's also not guaranteed).

Makes sense!

PR #734, opened 2026-02-06
Add the response message and error to LLM server.

RobinChiu:
v2 running the swe bench verified will have a lot of LimitsExceeded error. Checking the traj.json found 
1. the message before the tool call error didn't add to messages and feed back to LLM server
2. the json exception error didn't put in the user message
It will cause the LLM server hard to fix it and keep trying the error response format.
After fixing it, the benchmark score is more reasonable and mini-extra inspect view is more clear. 
螢幕擷取畫面 2026-02-06 152126

### DIFF
--- a/src/minisweagent/config/benchmarks/swebench.yaml
+++ b/src/minisweagent/config/benchmarks/swebench.yaml
@@ -143,6 +143,7 @@ model:
     }
     {%- endif -%}
   format_error_template: |
+    {{error}}
     Tool call error. Every response needs to use the 'bash' tool at least once to execute commands.
 
     Call the bash tool with your command as the argument:
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -9,6 +9,7 @@ from typing import Any, Literal
 import litellm
 from pydantic import BaseModel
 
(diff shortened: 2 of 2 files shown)


klieret:
Oh, the missing `{{error}}` is definitely a mistake on my part. 

I feel less sure about the other part though. I'm worried that adding the full incorrect response to the history will cause problems. For example, every tool call will need to be answered, so it will actually add not only that message but also additional messages that answer the toolcalls with empty output (else you get API errors).

klieret:
So I think if there's specific error cases, we should update the `error` text of the `FormatError` rather than appending the actual erroneous message that we didn't execute

klieret:
Somehow I can't push on this, so opened #735 where I built on your PR but added the change to all prompts. 

I rolled back the changes to the litellm model though as described above

RobinChiu:
Currently, when an Errorformat occurs, the LLM is not given the information(the incorrect response) to explaining why the error occurred. I suspect the LLM will be unable to correct the error. And the traj without the incorrect response information to debug it. Maybe we should consider at least record the error part information only. I found the same issue in the 
https://github.com/SWE-agent/mini-swe-agent/pull/728.

PR #740, opened 2026-02-11
litellm default timeout is 600s. Change it to 60s for swebench.

RobinChiu:
Set the litellm timeout to 60s when run the swebench.


klieret:
Having it at 600/the default setting seems better to me. People often run things overnight. Would be much more annoying to abort per false positive than to just be stuck

RobinChiu:
I saw there is 60s in the 
https://github.com/SWE-agent/mini-swe-agent/blob/main/src/minisweagent/models/openrouter_model.py#L77
Should we keep the same timeout number for different model?

PR #747, opened 2026-02-18
Fix: Update SwerexModalEnvironment for v2 Environment protocol

arbipher:
The code change and PR were initialized by Claude, but I tested it and checked it on my previous workload.
I have also read the Contributing, including `pre-commit` and `pytest`. Some code is learnby from `swerex_docker.py`.

---

Add missing `serialize()` method, matching the pattern used by `DockerEnvironment` and `LocalEnvironment`
Update `execute()` to accept `action: dict` instead of `command: str`, aligning with the v2 Environment protocol where actions are passed as dicts with a "command" key
Add missing `_check_finished()` method to detect `COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT` and raise `Submitted`, matching all other environment implementations
Add tests for all three fixes

### DIFF
--- a/src/minisweagent/environments/extra/swerex_modal.py
+++ b/src/minisweagent/environments/extra/swerex_modal.py
@@ -5,6 +5,9 @@ from pydantic import BaseModel
 from swerex.deployment.modal import ModalDeployment
 from swerex.runtime.abstract import Command as RexCommand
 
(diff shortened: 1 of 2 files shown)


klieret:
Tsym for opening this!

klieret:
Released in v2.2.1 thanks again!

PR #751, opened 2026-02-20
Fixed a bug inside ContreeEnvironmentConfig.get_tag_by_image_url

delphikettle:
There was a bug inside get_tag_by_image_url, it produced an error on some images

### DIFF
--- a/src/minisweagent/environments/extra/contree.py
+++ b/src/minisweagent/environments/extra/contree.py
@@ -37,7 +37,7 @@ class ContreeEnvironmentConfig(BaseModel):
     In case of conflict with `env`, the `env` variables take precedence.
     """
     interpreter: list[str] = Field(default_factory=lambda: ["bash", "-c"])
-    """"""
+    """Interpreter to execute commands"""
     timeout: int = 30
     """Timeout for executing commands in the container."""
 


src/minisweagent/environments/extra/contree.py
```diff
@@ -36,8 +36,8 @@ class ContreeEnvironmentConfig(BaseModel):
     Variables are only forwarded if they are set in the host environment.
     In case of conflict with `env`, the `env` variables take precedence.
     """
-    interpreter: list[str] | None = Field(default=lambda: ["bash", "-c"])
-    """"""
+    interpreter: list[str] = Field(default_factory=lambda: ["bash", "-c"])
```
klieret, line 39: Can't we just do this:

```suggestion
interpreter: list[str] = ["bash", "-c"]
```

since this is pydantic it's not a "mutable default"

src/minisweagent/environments/extra/contree.py
```diff
@@ -156,10 +156,15 @@ def serialize(self) -> dict:
     @staticmethod
     def get_tag_by_image_url(url: str) -> str:
         url_parsed = urlparse(url)
-        url = url_parsed.path
+        if url_parsed.netloc:
+            url = url_parsed.path
+
         if ":" not in url:
             url += ":latest"
-        domain, url_path = url.split("/", 1)
```
klieret, line 162: so I guess the issue was that sometimes there's no `/` (if we just have a direct image name)?

src/minisweagent/environments/extra/contree.py
klieret, line 32: also here and in `forward_env`, we can just do `={}` and `=[]`

klieret:
Let me merge this & release just so it's fixed, but would appreciate a PR with the simplifications :)

klieret:
Also, could you add a test for whatever this was fixing?

klieret:
released in 2.2.3

delphikettle:
@klieret, thank you so much for the review and merge! I will provide PR with fixes later with all requested fixes and tests later.

PR #763, opened 2026-03-01
Add default-install Bedrock support for AWS_BEARER_TOKEN_BEDROCK

### LINKED ISSUE #762: Bedrock models don't work out of the box with AWS_BEARER_TOKEN_BEDROCK
## Summary

`mini-swe-agent` currently cannot use Bedrock models out of the box when authentication is provided via `AWS_BEARER_TOKEN_BEDROCK`.

The project routes Bedrock models through LiteLLM, but the default install does not include `boto3`. Since LiteLLM's Bedrock path depends on the AWS SDK, Bedrock usage fails at runtime with a missing dependency error even when the auth token is correctly set.

## Problem

Today:

- Bedrock models can be selected through the existing LiteLLM-backed model classes
- `AWS_BEARER_TOKEN_BEDROCK` can be stored via `mini-extra config set ...`
- but `boto3` is only installed in the `modal` extra, not in the default package install

madmtang:
Fixes #762 

## Summary

This makes Bedrock work in a default `mini-swe-agent` install when using LiteLLM with `AWS_BEARER_TOKEN_BEDROCK`.

## Changes

- add `boto3>=1.39.0` to base dependencies
- remove `boto3` from the `modal` extra
- add a clearer Bedrock-specific error for missing `boto3` / `botocore` in:
  - `litellm`
  - `litellm_textbased`
  - `litellm_response`
- add test coverage for the new error handling and setup text

## Why

Without this, Bedrock can fail on a normal install due to missing AWS SDK deps, even if the user has configured auth correctly.

This keeps the current architecture unchanged:
- no new model class
- no new config fields
- still uses the existing LiteLLM-backed paths

### DIFF
--- a/docs/models/quickstart.md
+++ b/docs/models/quickstart.md
@@ -26,7 +26,6 @@ There are several ways to set your API keys:
     ALEPHALPHA_API_KEY
     ANTHROPIC_API_KEY
     ANYSCALE_API_KEY
-    AWS_BEARER_TOKEN_BEDROCK
     AZURE_AI_API_KEY
     AZURE_API_KEY
     AZURE_OPENAI_API_KEY
@@ -74,7 +73,6 @@ There are several ways to set your API keys:
     ```
 
     In addition, Portkey models use the `PORTKEY_API_KEY` environment variable.
-    For Bedrock API-key auth, use `AWS_BEARER_TOKEN_BEDROCK`. Standard AWS credentials also work if your environment already provides them.
 
 ## Selecting a model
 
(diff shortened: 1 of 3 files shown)


klieret:
Hi @madmtang , thanks for this, but in order to keep `mini` minimal, I think we shouldn't do this. This seems to be a very consistent thing, i.e., if anyone tries to run with bedrock, they'll see a warning about the dependency issue and will fix it before they deploy it. So I think this is the user's responsibility (seems like `litellm` also treats it this way).

PR #764, opened 2026-03-01
fix(modal): Add missing exception_info to SwerexModalEnvironment output

AlienKevin:
## Summary
- After recent v2 changes, the `execute()` method in `SwerexModalEnvironment` was missing the `exception_info` key in its return dict, unlike all other environment implementations (docker, local, singularity, swerex_docker). This caused a Jinja2 `UndefinedError` when the observation template accessed `output.exception_info` with `StrictUndefined`, crashing the agent on every step.
- Also adds `try/except` around the execute call to match `swerex_docker`, so execution errors are returned as structured output instead of propagating as unhandled exceptions.

## Test plan
- [x] Verified fix by running `mini-extra swebench` with Modal environment on SWE-smith-rs tasks — all instances now complete successfully instead of failing with `UndefinedError`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


klieret:
Your commit is now included in  2.2.6. Thanks again, I truly appreciate these bugfixes ❤️

PR #765, opened 2026-03-02
 Fixed ContreeEnvironment for new contree-sdk version

delphikettle:


delphikettle:
Hi, @klieret! Following up [previous PR](https://github.com/SWE-agent/mini-swe-agent/pull/751) I've made this PR. Since now this integration is based on newer version of `contree-sdk`, `get_tag_by_image_url` is no longer needed and thus there's no need for `get_tag_by_image_url` tests as different cases are already covered in the `contree-sdk` tests

klieret:
great, thanks! Is this a fix or will the current version still work? (just to check if I should release immediately or later on)

delphikettle:
Current version will still work, but we are deprecating the method, that was used before, so it's important for me to make this contribution, so any possible new users would use new more reliable API

There's no rush to release it immediately though

klieret:
Awesome, thanks!

PR #776, opened 2026-03-10
add run args for podman

Chibukach:
On servers with a rootless user running podman, this maps the current rootless user’s UID:GID to the same values in the container.


klieret:
But you could just add that to run args, right? If you want to describe these changes in the documentation, I'd merge a doc PR, but I don't think we need to hardcode that, because then it also is impossible for users to remove the flags

PR #778, opened 2026-03-10
use custom image path

Chibukach:
The agent pulls from swebench in dockerhub using the latest image tag. This allows the user to provide their own image registry, repository and tag as an environment variable.


klieret:
Hi, thanks for the suggestion, but this line of code is only a stopgap measure if the image name is not in the huggingface file. 
If you need custom domains, just create a custom dataset with the image names in it

PR #781, opened 2026-03-11
Fix TypeError when total_tokens is None in PortkeyModel

hobostay:
## Summary

Fix a bug where `total_tokens` being None would cause a TypeError in
cost calculation. This can happen with some Portkey API responses.

## Changes

- **portkey_model.py**: Add check for None total_tokens and calculate from prompt_tokens + completion_tokens
- **test_portkey_model.py**: Add test case for total_tokens=None scenario

## Testing

Added test `test_portkey_model_total_tokens_none` that simulates the
bug scenario and verifies the fix works correctly.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/minisweagent/models/portkey_model.py
+++ b/src/minisweagent/models/portkey_model.py
@@ -167,6 +167,11 @@ class PortkeyModel:
                 f"Completion tokens are None for model {self.config.model_name}. Setting to 0. Full response: {response_for_cost_calc.model_dump()}"
             )
             completion_tokens = 0
+        if total_tokens is None:
+            logger.warning(
+                f"Total tokens are None for model {self.config.model_name}. Setting to sum of prompt and completion tokens. Full response: {response_for_cost_calc.model_dump()}"
+            )
+            total_tokens = prompt_tokens + completion_tokens
         if total_tokens - prompt_tokens - completion_tokens != 0:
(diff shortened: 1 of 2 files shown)


klieret:
wouldn't prompt and completion tokens then also be None and it would still fail?

hobostay:
@klieret Thanks for the review! That's a valid concern.

You're right that in theory prompt_tokens or completion_tokens could also be None. However, due to the execution order, they are handled before total_tokens calculation (lines 158-169), so they are guaranteed to be non-None when we reach line 174.

I have now added a defensive get_token helper function to handle this edge case explicitly, making the code more robust against unexpected API responses.

klieret:
I don't like that this could end up with total tokens = 0, which is dangerous, because then a user could set a cost limit that is never reached and the agent never terminates.

Basically right now the `if total_tokens - prompt_tokens - completion_tokens != 0:` should ensure that two out of the three things are not None. 

This change could allow to have only Nones in the response. I think at this point this is too broken to recover, I'd rather have the program fail than to risk reporting only 0s for cost. 

I'm going to close this for now, because we have already too many open PRs, but if you have some data aobut how often this happens in practice, I'd be willing to reconsider.

PR #805, opened 2026-04-03
Handle tool calls split across chat choices

### LINKED ISSUE #804: Tool-call parsing fails when assistant text and tool calls are split across chat choices
## Summary

`mini-swe-agent` tool-call parsing can fail when a chat-completions provider returns assistant text and tool calls in separate `choices` within the same response.

In the failing case I observed, the first choice contained assistant text and a later choice contained the `bash` tool call. Because `mini-swe-agent` only inspected `response.choices[0]`, it raised a format error saying no tool calls were found.

## Observed behavior

A valid tool call is returned by the provider, but `mini-swe-agent` fails with an error like:

```text
Tool call error:

<error>
No tool calls found in the response. Every response MUST include at least one tool call.
</error>
```

## Expected behavior
If a valid tool call is present anywhere in the returned assistant choices, mini-swe-agent should parse it instead of assuming everything is always in choices[0].

## Notes
I hit this with GitHub Copilot + Claude Sonnet 4.6, where the response shape looked like:

- one assistant choice with text content

- another assistant choice with the tool call

This appears to be a valid provider/adaptor response shape that mini-swe-agent should tolerate.

## Related context
I initially opened a related LiteLLM issue while investigating:

- [BerriAI/litellm#24565](https://github.com/BerriAI/litellm/issues/24565)

After digging further, the immediate failure appears to be in mini-swe-agent's response parsing assumptions rather than in the wrapper layer itself.

simpleusr:
Fixes #804

## Summary

This PR fixes tool-call parsing when a chat-completions provider returns assistant text and tool calls in separate `choices` within the same response.

`mini-swe-agent` currently assumes the relevant assistant message is always in `response.choices[0]`. In practice, some provider integrations can return assistant text in one choice and the tool call in another, which causes tool-call parsing to fail even though a valid tool call was returned.

## Changes

- merge assistant content and `tool_calls` across returned chat choices before parsing
- update `LitellmModel` to read merged message/tool-call data instead of only `choices[0]`
- update `PortkeyModel` similarly for consistency
- make tool-call parsing accept both object-style and dict-style tool call entries
- add regression tests for split-choice tool-call responses
- sanitize task-specific fixture names in tests

## Why this helps

This makes tool-call parsing more robust across provider adapters without changing the normal single-choice path.

## Verification

I added regression tests covering:
- assistant text in one choice and tool calls in another
- dict-style tool calls after message normalization

## Reviewer Checklist
- confirm merging across chat choices is acceptable for tool-call models
- confirm object-style and dict-style tool-call parsing both look reasonable
- confirm the added regressions cover the intended response shape without overfitting to one provider

### DIFF
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -13,6 +13,7 @@ from minisweagent.models import GLOBAL_MODEL_STATS
 from minisweagent.models.utils.actions_toolcall import (
     BASH_TOOL,
     format_toolcall_observation_messages,
+    merge_choice_messages,
     parse_toolcall_actions,
 )
 from minisweagent.models.utils.anthropic_utils import _reorder_anthropic_thinking_blocks
@@ -83,7 +84,7 @@ class LitellmModel:
                 response = self._query(self._prepare_messages_for_api(messages), **kwargs)
         cost_output = self._calculate_cost(response)
         GLOBAL_MODEL_STATS.add(cost_output["cost"])
-        message = response.choices[0].message.model_dump()
+        message = merge_choice_messages(response.choices)
         message["extra"] = {
             "actions": self._parse_actions(response),
             "response": response.model_dump(),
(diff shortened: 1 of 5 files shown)


simpleusr:
Tested against a real failing provider response (github_copilot/claude-sonnet-4.6) mentioned in [original litellm issue](https://github.com/BerriAI/litellm/issues/24565)

klieret:
Thanks for the PR!

As I commented in #804 this really seems a model issue, not something that's mini's fault. So I think we shouldn't preemtively complicate our logic for this.

PR #807, opened 2026-04-04
fix: raise ContainerNotRunning when container dies after timeout

### LINKED ISSUE #803: container_timeout expiry causes silent failure loop instead of clean agent exit
### Describe the bug

Bug Description

  When a Docker container exits after container_timeout (implemented as sleep <duration>), the agent process continues running and wastes API calls until step_limit or cost_limit is exhausted.

  Root Cause

  container_timeout is implemented by starting the container with sleep <duration>. Once the sleep exits, the container is stopped and removed (--rm). Subsequent docker exec calls fail, but
  execute() catches the exception and returns it as a normal observation (returncode=-1, exception_info set) without raising. The model sees error output but has no way to know the container is
  gone, so it keeps issuing commands until limits are hit.

  Relevant code path in environments/docker.py:

  except Exception as e:
      output = {
          "output": raw_output,
          "returncode": -1,
          "exception_info": f"An error occurred while executing the command: {e}",
          ...
      }
  # _check_finished only checks for COMPLETE_TASK_AND_SUBMIT_FINAL_OUTPUT
  self._check_finished(output)
  return output  # silently swallowed

  Impact

  - In benchmarks with step_limit=500, a task whose container dies at step 200 will continue for another 300 steps, burning ~300 unnecessary API calls and the associated cost.
  - The trajectory is never saved with a proper exit status.

  Proposed Fix

  1. Add a ContainerNotRunning exception subclassing InterruptAgentFlow in exceptions.py:

  class ContainerNotRunning(InterruptAgentFlow):
      """Raised when the Docker container is no longer running."""

  2. In execute(), detect container-dead markers in both the normal return path and the exception path:

  _CONTAINER_DEAD_MARKERS = ("No such container", "is not running", "no such container")

  # after subprocess.run returns
  if result.returncode != 0 and any(m in result.stdout for m in _CONTAINER_DEAD_MARKERS):
      raise ContainerNotRunning(
          {"role": "exit", "content": result.stdout,
           "extra": {"exit_status": "ContainerNotRunning", "submission": ""}}
      )

  # in except block
  if any(m in str(e) for m in _CONTAINER_DEAD_MARKERS):
      raise ContainerNotRunning(
          {"role": "exit", "content": str(e),
           "extra": {"exit_status": "ContainerNotRunning", "submission": ""}}
      )
 Because ContainerNotRunning is an InterruptAgentFlow subclass, the agent's run() loop catches it, records the exit message, and the finally: self.save() block saves the trajectory — exactly the
  same clean shutdown path used by LimitsExceeded.

  Result After Fix

  - Agent terminates immediately when the container dies   
  - Trajectory saved with exit_status: "ContainerNotRunning"
  - Counts as a failed instance (not submitted), no wasted API calls

GopalGB:
## Summary

- Adds `ContainerNotRunning` exception (subclass of `InterruptAgentFlow`) that terminates the agent loop cleanly when the Docker container dies after `container_timeout` expires
- Detects container-dead markers (`"No such container"`, `"is not running"`) in both subprocess output and exception messages
- Prevents the agent from silently burning API calls after the container is gone

## Problem

When `container_timeout` expires, the container's `sleep` process exits and Docker removes it (`--rm`). Subsequent `docker exec` calls fail, but `execute()` catches the exception and returns it as a normal observation. The model keeps issuing commands until `step_limit` or `cost_limit` is exhausted, wasting API calls and cost.

## Solution

`ContainerNotRunning` extends `InterruptAgentFlow`, so the agent's existing `run()` loop handles it identically to `LimitsExceeded` or `Submitted`: it adds the exit message, saves the trajectory, and stops.

## Test plan

- [x] Unit test: `ContainerNotRunning` raised when subprocess returns dead-container output (non-zero returncode)
- [x] Unit test: `ContainerNotRunning` raised when subprocess raises exception with dead-container message
- [x] Integration test: `ContainerNotRunning` raised after real `container_timeout` expires (requires Docker, marked `@pytest.mark.slow`)
- [x] All existing tests pass

Fixes #803

### DIFF
--- a/src/minisweagent/environments/docker.py
+++ b/src/minisweagent/environments/docker.py
@@ -8,8 +8,10 @@ from typing import Any
 
 from pydantic import BaseModel
 
-from minisweagent.exceptions import Submitted
+from minisweagent.exceptions import ContainerNotRunning, Submitted
 from minisweagent.utils.serialize import recursive_merge
+
+_CONTAINER_DEAD_MARKERS = ("No such container", "is not running", "no such container")
 
 
 class DockerEnvironmentConfig(BaseModel):
@@ -123,11 +125,31 @@ class DockerEnvironment:
                 stderr=subprocess.STDOUT,
             )
             output = {"output": result.stdout, "returncode": result.returncode, "exception_info": ""}
+            if result.returncode != 0 and any(marker in result.stdout for marker in _CONTAINER_DEAD_MARKERS):
+                raise ContainerNotRunning(
+                    {
+                        "role": "exit",
+                        "content": result.stdout,
+                        "extra": {"exit_status": "ContainerNotRunning", "submission": ""},
+                    }
+                )
+        except ContainerNotRunning:
+            raise
         except Exception as e:
             raw_output = getattr(e, "output", None)
             raw_output = (
                 raw_output.decode("utf-8", errors="replace") if isinstance(raw_output, bytes) else (raw_output or "")
             )
+            if any(marker in str(e) for marker in _CONTAINER_DEAD_MARKERS) or any(
+                marker in raw_output for marker in _CONTAINER_DEAD_MARKERS
+            ):
+                raise ContainerNotRunning(
+                    {
+                        "role": "exit",
+                        "content": raw_output or str(e),
+                        "extra": {"exit_status": "ContainerNotRunning", "submission": ""},
+                    }
+                )
             output = {
                 "output": raw_output,
                 "returncode": -1,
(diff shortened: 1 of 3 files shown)


GopalGB:
Friendly ping — this PR has been clean (mergeStateStatus: CLEAN, pre-commit.ci passed) for a month. The fix raises `ContainerNotRunning` instead of a misleading exception when the container dies after timeout, surfacing the real failure mode to operators. Happy to address any review feedback. cc maintainers.

klieret:
Hi, thanks for opening this, I'm currently catching up with issues & PRs, we were in a big grind to release programbench. Will take a look soon.

klieret:
Thanks for opening this @GopalGB 

I think the biggest problem is that the markers could also appear in normal output. E.g., if mini-swe-agent itself tries to run docker in a docker in docker setup, or even is just handling code that contains these words.  

I've just merged a different fix, #832

PR #810, opened 2026-04-07
docs: remove model-specific selection recommendations

kiankyars:
## Summary
- remove the "Which model to use" recommendation block from `docs/quickstart.md`
- remove the hard-coded "Popular models" list from `docs/models/quickstart.md`
- keep model setup guidance provider-agnostic and point readers to current provider/model catalogs

## Test plan
- [x] docs-only change
- [x] verified markdown renders cleanly in diff

Made with [Cursor](https://cursor.com)


kiankyars:
@klieret I made this PR because all of the models listed are  no longer relevant.

PR #818, opened 2026-04-19
[fix] preserve explicit zero cost limit

### LINKED ISSUE #813: Bug: --cost-limit 0 silently ignored (falsy float defeats "or UNSET" check)
## `--cost-limit 0` is silently ignored

The `--cost-limit` help text says "Set to 0 to disable", but passing `--cost-limit 0` has no effect. The default `cost_limit: 3.0` from `mini.yaml` always wins.

The problem is in `run/mini.py:80` and `run/benchmarks/swebench_single.py:75`:

```python
"cost_limit": cost_limit or UNSET,
```

Typer parses `--cost-limit 0` into `float(0.0)`, which is falsy in Python, so `0.0 or UNSET` evaluates to `UNSET` and the CLI value is discarded.

### Fix

```diff
- "cost_limit": cost_limit or UNSET,
+ "cost_limit": cost_limit if cost_limit is not None else UNSET,
```

Same change in both files.

kolygri:
## What changed
- preserve an explicit `--cost-limit 0` in `mini` and `swebench-single` instead of treating it as an unset value
- add focused regression coverage for both entrypoints so `cost_limit=0` survives config merging

## Why it changed
- the CLI help says `0` disables the cost limit, but the runners were dropping that value before agent config was built

## User or developer impact
- users can now pass `--cost-limit 0` and reliably disable the per-run cost cap from the CLI
- the new regression tests make future regressions in this config path much harder to reintroduce

## Root cause
- both runners used `cost_limit or UNSET`, and Python treats `0.0` as falsy, so the explicit CLI value was discarded and the config default won instead

## Validation performed
- `uv run pytest tests/run/test_cli_integration.py -k cost_limit_zero_is_preserved`
- `uv run python` isolated verification of the `swebench_single` config path with `datasets` stubbed
- `python -m compileall src/minisweagent/run/mini.py src/minisweagent/run/benchmarks/swebench_single.py tests/run/test_cli_integration.py tests/run/test_swebench_single.py`

## Review focus
- confirm the CLI merge behavior now distinguishes `None` from `0`
- confirm both entrypoints still preserve existing defaults when `--cost-limit` is omitted
- confirm the new tests cover the intended regression without changing unrelated runner behavior
- note that full `tests/run/test_swebench_single.py` collection is still blocked locally because this Python build lacks `_lzma` for `datasets`

Closes #813.

### DIFF
--- a/src/minisweagent/run/benchmarks/swebench_single.py
+++ b/src/minisweagent/run/benchmarks/swebench_single.py
@@ -72,7 +72,7 @@ def main(
         "agent": {
             "agent_class": agent_class or UNSET,
             "mode": "yolo" if yolo else UNSET,
-            "cost_limit": cost_limit or UNSET,
+            "cost_limit": cost_limit if cost_limit is not None else UNSET,
             "confirm_exit": False if exit_immediately else UNSET,
             "output_path": output or UNSET,
         },
--- a/src/minisweagent/run/mini.py
+++ b/src/minisweagent/run/mini.py
@@ -77,7 +77,7 @@ def main(
         "agent": {
             "agent_class": agent_class or UNSET,
             "mode": "yolo" if yolo else UNSET,
-            "cost_limit": cost_limit or UNSET,
+            "cost_limit": cost_limit if cost_limit is not None else UNSET,
             "confirm_exit": False if exit_immediately else UNSET,
             "output_path": output or UNSET,
         },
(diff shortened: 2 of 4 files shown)


klieret (approved):
Thank you, this was a subtle catch. Appreciated!

klieret:
superseded by #825 which contains the unittest fix

PR #830, opened 2026-05-20
feat: pluggable compaction strategies for context window management

vamshinannuta521:
## Summary

- Adds a `minisweagent.compaction` package with provider-scoped, pluggable context compaction strategies
- Introduces `CompactionStrategy` protocol in `minisweagent.__init__` alongside existing `Model`, `Environment`, `Agent` protocols
- Wires compaction into `LitellmModel` and `OpenRouterModel` via `_prepare_messages_for_api()` — only the API-bound view is compacted; `self.messages` (the full trajectory) is never mutated
- Adds `--compaction` CLI flag to `mini` and `model.compaction` YAML key

## Strategies included

| Name | Description |
|---|---|
| `none` | Default no-op pass-through |
| `generic/sliding_window` | Keep last N steps (system + first user always preserved) |
| `generic/summarize` | LLM-based summarization of older steps |
| `anthropic/summarize` | Same, defaults to `claude-haiku-4-5-20251001` |

## Usage

```bash
# CLI
mini --compaction generic/sliding_window -c model.compaction_kwargs.keep_last_n_steps=8 -t "fix the bug"
mini --compaction anthropic/summarize -c model.compaction_kwargs.keep_last_n_steps=3 -t "fix the bug"

# YAML
# model:
#   compaction: generic/sliding_window
#   compaction_kwargs:
#     keep_last_n_steps: 8
```

## Test plan

- [ ] `MSWEA_SILENT_STARTUP=1 .venv/bin/python -m pytest tests/ -q` — all previously passing tests still pass
- [ ] Smoke test: `get_compaction_strategy("generic/sliding_window", keep_last_n_steps=1).compact(msgs)` returns correct window
- [ ] `isinstance(strategy, CompactionStrategy)` returns `True` for all built-in strategies

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/minisweagent/__init__.py
+++ b/src/minisweagent/__init__.py
@@ -12,7 +12,7 @@ __version__ = "2.2.8"
 
 import os
 from pathlib import Path
-from typing import Any, Protocol
+from typing import Any, Protocol, runtime_checkable
 
 import dotenv
 from platformdirs import user_config_dir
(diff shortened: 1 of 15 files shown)


klieret:
Hi thanks, appreciate the PR, happy to link to this in a fork or add it to some documentation about how to implement this.

However, with this being focused on being a minimal coding agent and with seeing very few (if any) benchmarks where we actually hit the context limit, I'd rather not merge this at this point in time.

It would make things more complex and require more tuning of specific prompts, so I think this is a bit out of scope here.

But I really appreciate this, anyway, thanks!

PR #831, opened 2026-05-20
docs: add Tuning Engines LiteLLM example

cerebrixos:
## Summary

Adds a docs-only Tuning Engines example to the model setup quickstart. The example uses the existing LiteLLM path with an OpenAI-compatible API base and a Tuning Engines inference key.

## Testing

Not run; docs-only change.


klieret:
hi, thanks for opening this, but don't want to clutter the docs too much at this point

PR #837, opened 2026-05-24
docs: fix broken banner link

Chesars:
The banner was adding latest twice, so the link pointed to: `/latest/latest/usage/programbench/`
Use a docs-relative link instead so MkDocs builds the correct URL.


klieret:
thanks!

PR #847, opened 2026-06-01
Chore - DevOps automation to clean repo of stale branches over 90 days old

schneidergithub:
This is a typical devops workflow to help keep the repo clean of stale branches.  I noticed there are some old branches 9 months old that can probably be removed, and this automates that whole process going forward.


klieret:
i think we're fine without it rn

PR #850, opened 2026-06-04
Don't prompt for new limits when stdin is non-interactive

### LINKED ISSUE #849: InteractiveAgent crashes with EOFError on limit breach when stdin is non-interactive
## Summary

`InteractiveAgent` crashes with `EOFError` when a step/cost limit is hit during an
unattended run (no interactive terminal), instead of stopping cleanly at the limit.

## Details

`InteractiveAgent.query()` handles `LimitsExceeded` by prompting for new limits:

```python
except LimitsExceeded:
    ...
    self.config.step_limit = int(input("New step limit: "))
    self.config.cost_limit = float(input("New cost limit: "))
```

This assumes a human is at a terminal. When stdin is not interactive — redirected
from `/dev/null` in CI, a batch harness, or a sandbox (for example the way `pier`
launches `mini-swe-agent --yolo` inside a container) — `input()` raises `EOFError`.
That exception is uncaught, so it propagates out of `run()` and crashes the whole
run. A run that should have ended with a clean `LimitsExceeded` exit (trajectory
saved) is instead recorded as an `EOFError` agent error.

`--yolo` does not avoid this: it only suppresses *command* confirmation, not the
limit-extension prompt.

### Repro

```bash
mini --yolo -m <model> -t "<task>" -c agent.cost_limit=0.01 < /dev/null
# -> EOFError: EOF when reading a line
```

(any non-interactive stdin reproduces it; `< /dev/null` is the simplest)

## Second, related bug: `TimeExceeded` prompts forever

`TimeExceeded` subclasses `LimitsExceeded`, so it hits the same handler — but
raising the step/cost limits can't satisfy a wall-clock budget. After the prompt,
the next `query()` re-checks the clock and raises `TimeExceeded` again, re-prompting
in a loop. With a real terminal this loops forever; with non-interactive stdin it
`EOFError`s as above.

## Expected behaviour

When there's no interactive terminal, a breached limit should stop the run cleanly
with a `LimitsExceeded` exit status and a saved trajectory (the same clean stop a
non-interactive agent already produces). `TimeExceeded` should always stop rather
than prompt.

## Fix

I have a fix + tests ready and will open a PR referencing this issue. It gates the
prompt on `sys.stdin.isatty()` (re-raising otherwise) and always re-raises
`TimeExceeded`; interactive behaviour is unchanged.

Entrpi:
## Problem

`InteractiveAgent.query()` handles a breached limit by prompting for new step and
cost limits via `input()`:

```python
except LimitsExceeded:
    ...
    self.config.step_limit = int(input("New step limit: "))
    self.config.cost_limit = float(input("New cost limit: "))
```

That assumes a human is at a terminal. In an **unattended run** there isn't one,
and the behaviour breaks in two ways:

1. **Crash on non-interactive stdin.** When stdin is redirected from `/dev/null`
   (CI, a batch harness, a sandbox — e.g. the way `pier` launches `mini-swe-agent
   --yolo` inside a container), `input()` raises `EOFError`. That uncaught
   exception propagates out of `run()`, so what should be a clean, trajectory-
   preserving budget stop instead crashes the whole trial and is recorded as an
   `EOFError` agent error rather than a normal `LimitsExceeded` exit.

   Note `--yolo` does **not** avoid this: it only suppresses *command*
   confirmation; the limit-extension prompt still fires.

2. **Infinite prompt loop on a time limit.** `TimeExceeded` subclasses
   `LimitsExceeded`, so it hits the same handler — but raising the step/cost
   limits can't satisfy a wall-clock budget. The next `query()` re-checks the
   clock and raises `TimeExceeded` again, re-prompting forever (even with a real
   terminal).

### Repro

```bash
mini --yolo -m <model> -t "<task>" -c agent.cost_limit=0.01 < /dev/null
# -> EOFError: EOF when reading a line   (run crashes instead of stopping at the limit)
```

## Fix

In `InteractiveAgent.query()`:

- Only prompt for new limits when an interactive terminal is actually available
  (`sys.stdin.isatty()`). Otherwise re-raise `LimitsExceeded`, so
  `DefaultAgent.run()` records a normal `LimitsExceeded` exit and saves the
  trajectory — the same clean stop a non-interactive agent already gets.
- Always re-raise `TimeExceeded` (it can't be lifted by raising step/cost limits),
  avoiding the infinite-prompt loop.

A small `_stdin_is_interactive()` helper also guards against `ValueError`/`OSError`
from a closed stdin.

Behaviour with a real terminal is unchanged: an interactive user still gets the
"New step limit / New cost limit" prompts on a step/cost breach exactly as before.

## Tests

- Add `test_limits_exceeded_non_interactive_stops_cleanly` — `--yolo` with no TTY:
  the run stops with a `LimitsExceeded` exit status and `input()` is never called.
- Add `test_time_exceeded_never_prompts` — even with a TTY, a wall-clock breach
  stops cleanly without prompting.
- The two existing limit-continuation tests already mock `input()` (i.e. they
  simulate a human at a terminal); they're updated to report an interactive TTY so
  they continue to exercise the prompt path.

`tests/agents/test_interactive.py` (108) and `tests/agents/test_default.py` (45)
pass; `ruff check` / `ruff format --check` clean.

Fixes #849.

### DIFF
--- a/src/minisweagent/agents/interactive.py
+++ b/src/minisweagent/agents/interactive.py
@@ -7,6 +7,7 @@ There are three modes:
 """
 
 import re
+import sys
 from typing import Literal, NoReturn
 
 from rich.console import Console
@@ -14,7 +15,7 @@ from rich.rule import Rule
 
 from minisweagent.agents.default import AgentConfig, DefaultAgent
 from minisweagent.agents.utils.prompt_user import _multiline_prompt, prompt_session
-from minisweagent.exceptions import LimitsExceeded, Submitted, UserInterruption
+from minisweagent.exceptions import LimitsExceeded, Submitted, TimeExceeded, UserInterruption
 from minisweagent.models.utils.content_string import get_content_string
 
 console = Console(highlight=False)
@@ -71,7 +72,19 @@ class InteractiveAgent(DefaultAgent):
         try:
             with console.status("Waiting for the LM to respond..."):
                 return super().query()
+        except TimeExceeded:
+            # A wall-clock limit can't be lifted by raising the step/cost limits
+            # (the next query re-checks the clock and raises again), so prompting
+            # would loop forever. Always stop cleanly instead.
+            raise
         except LimitsExceeded:
+            if not self._stdin_is_interactive():
+                # No terminal to prompt for new limits -- e.g. an unattended
+                # `--yolo` run, or any run with stdin redirected from /dev/null
+                # (CI, a sandbox). Stop cleanly so the trajectory is saved with a
+                # LimitsExceeded exit status, instead of crashing on EOFError when
+                # reading input.
+                raise
             console.print(
                 f"Limits exceeded. Limits: {self.config.step_limit} steps, ${self.config.cost_limit}.\n"
                 f"Current spend: {self.n_calls} steps, ${self.cost:.2f}."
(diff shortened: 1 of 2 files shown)


klieret (approved):

klieret:
Thank you, appreciate this. Will make it so that user can override time limit.

PR #821, opened 2026-04-29
fix: persist full model response in extra dict on FormatError (#784)

VittoriaLanzo:
When `_parse_actions()` raises a `FormatError`, the trajectory log recorded the error marker but not the raw model response that caused it. The structured response object — tool calls, choices, finish reason — was not preserved, leaving nothing to inspect when debugging format failures.

Closes #784

This PR wraps each wrapper's parse call site in `query()` across all seven model files. On `FormatError`, the response is serialized into `e.messages[0]["extra"]["response"]` before re-raising. If `model_dump()` fails on a non-serializable field, the fallback is `repr(response)`, so `extra["response"]` is always populated. The `FormatError` propagates identically — only the `extra` dict gains a key. No public signatures change.

**Before / after (illustrative)**

```python
# Before — extra on FormatError
{"interrupt_type": "FormatError"}

# After — extra on FormatError
{"interrupt_type": "FormatError", "response": {"id": "chatcmpl-...", "choices": [...], ...}}
```

On the success path, the `try/except` block is never entered and `extra` is unchanged.

The new test file has 16 tests covering all 7 wrappers and the `repr`-fallback path. Each test fails on pre-fix code.

---

<details>
<summary>Pipeline diagram</summary>

```mermaid
flowchart TD
  main(["main-thread orchestrator"])

  subgraph C1 ["layer-1 · planning"]
    a1a["pattern-extractor"] --> a1b["coverage-engineer"]
    a1b --> a1c["contribution-planner"]
    a1c --> a1d["architect-reviewer"]
    a1d -->|objections| a1c
  end

  human1[["human gate · spec approval"]]

  subgraph C2 ["initial-fix"]
    a2["code-author · impl + regression"]
  end

  subgraph C3 ["quality-polish"]
    a3a["readability-editor"] --> a3b["verbosity-trimmer"]
  end

  subgraph C4 ["hygiene-check"]
    a4a["scope-linter"] --> a4b["ci-gate · pre-adversarial"]
  end

  subgraph C5 ["adversarial-loop"]
    atk["attack-pass · senior-reviewer + red-team"]
    rev["revision-pass · senior-engineer + test-author"]
    rrv["re-review-pass · red-team + cold-reviewer"]
    triv{{"triviality-check"}}
    gate["merge-gate"]
    atk --> rev --> rrv --> triv
    triv -->|NON_TRIVIAL| atk
    triv -->|TRIVIAL| gate
  end

  subgraph C6 ["pr-preparation"]
[...]

PR #851, opened 2026-06-04
Handle reasoning truncation distinctly from a missing tool call

Entrpi:
Builds on #821 (persist the full model response in `extra` on `FormatError`).

## Problem

When a tool-calling model returns a turn with **no tool call**, the retry tells it
*"No tool calls found in the response. Every response MUST include at least one tool call."*
But that message is often **wrong**: a reasoning model can spend its entire `max_tokens`
completion budget on `reasoning_content` and never reach the tool-call emission. The response
comes back with `finish_reason="length"` (cut mid-output) or `"tool_calls"` with an empty
payload (cut right at the boundary) — i.e. it was **truncated**, not that it "forgot." Telling a
truncated model to "include a tool call next time" is the wrong corrective signal, and if it
keeps happening the run burns its whole budget in a bounce loop.

### How we found it

Running MiniMax-M3 through `mini-swe-agent`, the no-tool-call turns on a hard task had a perfectly
clean signature once the response was inspectable (thanks to #821): **every** one had
`completion_tokens == max_tokens` exactly, with the budget consumed by `reasoning_content`, while
every *successful* turn finished comfortably under the cap. They were all `max_tokens` truncations
— raising the cap eliminated them.

## Changes

**1. Tell the model it was truncated, not that it forgot** (`LitellmModel`)
- When the `FormatError` response is a truncation (detected via `finish_reason`: `length`, or
  `tool_calls` with an empty payload), swap the retry for a **truncation-aware** message (new
  `truncation_error_template`), flag `extra.truncated=True`, and log a distinct warning suggesting
  a higher `max_tokens`. Genuine no-tool-call turns (`finish_reason=stop`) keep the existing
  message. The response is still persisted on the error exactly as #821 does.

**2. Bound a runaway bounce loop** (`DefaultAgent`)
- New `max_consecutive_format_errors` (default **0 = off**): after that many no-tool-call /
  truncation turns in a row, stop cleanly with `exit_status=RepeatedFormatError` instead of
  looping until the budget is gone. A successful tool call resets the counter.

## Why

- **Correct corrective signal:** a truncated model is told it was cut off (be concise), not that
  it omitted a tool call — the latter doesn't address the actual cause.
- **No silent budget burn:** an opt-in cap turns a truncation bounce loop into a clean, scoreable
  stop instead of consuming the whole time/cost budget. (We saw a single task spend dozens of
  turns this way.)
- **Actionable signal:** `extra.truncated` + the warning point operators straight at `max_tokens`,
  which is the real fix for reasoning models.

## Testing

- Model-level: `finish_reason=length` and empty-`tool_calls`-payload produce the truncation
  message + `extra.truncated`; a genuine `finish_reason=stop` no-tool-call keeps the original
  message and is not flagged (the response is still persisted on the error in all cases).
- Agent-level: `max_consecutive_format_errors` terminates with `exit_status=RepeatedFormatError`;
  a success between errors resets the counter.
- `tests/models/test_litellm_model.py`, `tests/agents/test_default.py`,
  `tests/agents/test_interactive.py` pass; `ruff check` / `ruff format --check` clean.

## Notes

- Non-breaking: the retry/bounce behaviour is unchanged except that a truncation now gets an
  accurate message; `max_consecutive_format_errors` defaults to off.
- Scope: `LitellmModel` (the common litellm path). The same truncation check could be added to the
  `openrouter` / response-API toolcall models in a follow-up.

### DIFF
--- a/src/minisweagent/agents/default.py
+++ b/src/minisweagent/agents/default.py
@@ -12,7 +12,7 @@ from jinja2 import StrictUndefined, Template
 from pydantic import BaseModel
 
 from minisweagent import Environment, Model, __version__
-from minisweagent.exceptions import InterruptAgentFlow, LimitsExceeded, TimeExceeded
+from minisweagent.exceptions import FormatError, InterruptAgentFlow, LimitsExceeded, TimeExceeded
 from minisweagent.utils.serialize import recursive_merge
 
 
@@ -29,6 +29,10 @@ class AgentConfig(BaseModel):
     """Stop agent after exceeding (!) this cost."""
     wall_time_limit_seconds: int = 0
     """Stop agent after this many seconds of wall-clock time. 0 means no limit."""
+    max_consecutive_format_errors: int = 0
+    """Stop the agent after this many format errors in a row (e.g. repeated max_tokens truncations
+    that never produce a tool call). 0 means no limit. Guards against a model burning the whole
+    budget in a bounce loop; exits cleanly with exit_status=RepeatedFormatError."""
     output_path: Path | None = None
     """Save the trajectory to this path."""
 
@@ -44,6 +48,7 @@ class DefaultAgent:
         self.logger = logging.getLogger("agent")
         self.cost = 0.0
         self.n_calls = 0
+        self._consecutive_format_errors = 0
         self._start_time = time.time()
 
     def get_template_vars(self, **kwargs) -> dict:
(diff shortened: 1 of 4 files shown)


klieret:
I made some slightly different implementation choices and split it up in two PRs, but basically created #863 on top of your branch here so you still get co-authorship :)

Thank you for your contributions!

Entrpi:
Thanks, glad it helped! :)

PR #865, opened 2026-06-11
fix: kill local process group on timeout

### LINKED ISSUE #826: Bug: Agent-spawned scripts keep running after their bash command times out
I'm running mini-swe-agent on a coding benchmark. The agent will sometimes write a Python script to /tmp and run it as a normal foreground command `python3 /tmp/foo.py`. If the script is slow or has an infinite loop, the 30s environment timeout fires and the agent gets a "command timed out" observation and moves on.

But the Python process doesn't actually die. It keeps running, reparented to init, burning CPU. I ended up with a dozen of these after a single eval run, some going for over an hour after the agent had timed them out and forgotten about them.

The cause: LocalEnvironment.execute uses [subprocess.run(shell=True, timeout=...)](https://github.com/SWE-agent/mini-swe-agent/blob/bc85a45654e6348dcc6e4c5a40ad146ed0bb144d/src/minisweagent/environments/local.py#L28) without start_new_session=True. When the timeout fires, Python only SIGKILLs the immediate child — the bash shell. Bash's children (the python script) get orphaned to init and keep running.

I wasn't sure if this is an intentional design choice. I couldn't find an existing issue or comment about it.

---

Quick Repro:

```
python3 <<'EOF'
import subprocess, os, time
cmd = """
cat <<'INNER' > /tmp/leak_test.py
import time
while True:
    time.sleep(1)
INNER
python3 /tmp/leak_test.py
"""
print("Calling subprocess.run(...)")
t0 = time.time()
try:
    result = subprocess.run(
        cmd,
        shell=True,
        text=True,
        cwd=os.getcwd(),
        env=os.environ | {"PAGER": "cat"},
        timeout=3,
        encoding="utf-8",
        errors="replace",
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
    )
except subprocess.TimeoutExpired:
    print(f"Timed out after {time.time()-t0:.1f}s")
time.sleep(1)
EOF
```

Output:
```
Calling subprocess.run with full mini-swe-agent arg set...
Timed out after 3.0s
```

Then:
```
$ ps -eo pid,ppid,etime,cmd | grep leak_test.py | grep -v grep
 274081       1       00:08 python3 /tmp/leak_test.py
```

`subprocess.run` reported the timeout, but the `python3 /tmp/leak_test.py` grandchild is still running with PPID=1 (reparented to init).

Swing-G:
Fixes #826.

LocalEnvironment currently uses subprocess.run(shell=True, timeout=...), which can leave shell-spawned child processes running after the timeout kills only the shell.

This change runs local commands in a new process group on POSIX and kills that group on timeout. Windows keeps the existing subprocess.run path.

Tests:
- pytest tests/environments/test_local.py -q
- ruff check src/minisweagent/environments/local.py tests/environments/test_local.py
- ruff format --check src/minisweagent/environments/local.py tests/environments/test_local.py

### DIFF
--- a/src/minisweagent/environments/local.py
+++ b/src/minisweagent/environments/local.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import signal
 import subprocess
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


klieret:
Thank you very much, really appreciate it :) Refactored a bit.

in v2.4.1 now

PR #868, opened 2026-06-18
fix(env): run local commands through bash

### LINKED ISSUE #829: Bash tool launches Dash on Ubuntu environments
### Describe the bug

I believe the mini-swe-agent bash tool:

https://github.com/SWE-agent/mini-swe-agent/blob/bc85a45654e6348dcc6e4c5a40ad146ed0bb144d/src/minisweagent/environments/local.py#L28

Is launching the default shell, which isn't Bash on Ubuntu/Debian system. You can test it with this:

```bash
import subprocess

r = subprocess.run(
    "echo shell=$0; echo <(echo hello)",
    shell=True,
    text=True,
    capture_output=True,
)

print("stdout:", r.stdout)
print("stderr:", r.stderr)
print("returncode:", r.returncode)
```

You will get a syntax error. The fix is to add `executable="/bin/bash"` to `subprocess.run`. I haven't looked into the containerized runners, but it's likely to affect them too.

dannyward630:
## Summary

Fixes #829.

LocalEnvironment advertised and prompted bash commands, but executed them via `subprocess.Popen(..., shell=True)`. On Debian/Ubuntu, Python uses `/bin/sh` for `shell=True`, and `/bin/sh` is commonly `dash`, so bash-only syntax such as process substitution fails even though the tool is named `bash`.

This changes local command execution to use a configurable interpreter, defaulting to `bash -lc`, matching the Docker backend's default interpreter behavior more closely while preserving the existing timeout/process-group handling.

## Validation

- `.venv/bin/python -m pytest tests/environments/test_local.py`
- `.venv/bin/ruff check src/minisweagent/environments/local.py tests/environments/test_local.py`
- `git diff --check`

### DIFF
--- a/src/minisweagent/environments/local.py
+++ b/src/minisweagent/environments/local.py
@@ -28,7 +28,9 @@ class LocalEnvironment:
         command = action.get("command", "")
         cwd = cwd or self.config.cwd or os.getcwd()
         try:
-            result = _run(command, cwd, os.environ | self.config.env, timeout or self.config.timeout, self.config.interpreter)
+            result = _run(
+                command, cwd, os.environ | self.config.env, timeout or self.config.timeout, self.config.interpreter
+            )
             output = {"output": result.stdout, "returncode": result.returncode, "exception_info": ""}
         except Exception as e:
             raw_output = getattr(e, "output", None)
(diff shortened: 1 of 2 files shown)


dannyward630:
Closing in favor of #870, which implements the maintainer's shell-agnostic prompt/tool-description approach for #829.

klieret:
still thinking about this, your PR is also very clean (thx!), I'm just still wondering if just prefixing `bash -lc` to the Popen is fully equivalent or if there's any edge cases (I guess popen handles all the escaping etc, but is there more that could break)

klieret:
The one thing I don't like with this, is that this will probably not work out of the box on some systems...

PR #869, opened 2026-06-18
fix(env): stop or kill docker containers on cleanup

### LINKED ISSUE #797: Why use `||` between `stop` and `rm` when cleanup a docker environment?
### Describe the bug

Whenever I finish running mini-swe-agent, it consistently leaves behind a large number of exited Docker containers.
This appears to be because the commands within the `cleanup` method of the `DockerEnvironment` class use the logical OR operator (`||`) between `docker stop` and `docker rm`, resulting in the containers not being removed after their usage is complete.
Is this the intended behavior?
https://github.com/SWE-agent/mini-swe-agent/blob/58ad92fc1385be3f8f2326b7a42320bb3655e169/src/minisweagent/environments/docker.py#L153-L157

dannyward630:
## Summary

Fixes #797.

`DockerEnvironment.cleanup()` currently falls back from `docker stop` to `docker rm -f`. That makes cleanup force removal when `stop` fails, even though removal is already controlled by the configured `run_args` (`--rm` by default) and users may intentionally omit `--rm` to keep stopped containers for debugging.

This changes the fallback to `docker kill`, so cleanup attempts to terminate the container while leaving removal policy to Docker's `--rm` behavior or the user's `run_args` override.

## Validation

- `.venv/bin/python -m pytest tests/environments/test_docker.py -k 'config_defaults or cleanup_stops_or_kills'`
  - `1 passed, 2 skipped, 20 deselected` on this machine because Docker/Podman are not available
- `.venv/bin/ruff check src/minisweagent/environments/docker.py tests/environments/test_docker.py`
- `git diff --check`

### DIFF
--- a/src/minisweagent/environments/docker.py
+++ b/src/minisweagent/environments/docker.py
@@ -151,9 +151,9 @@ class DockerEnvironment:
             )
 
     def cleanup(self):
-        """Stop and remove the Docker container."""
+        """Stop the Docker container."""
         if getattr(self, "container_id", None) is not None:  # if init fails early, container_id might not be set
-            cmd = f"(timeout 60 {self.config.executable} stop {self.container_id} || {self.config.executable} rm -f {self.container_id}) >/dev/null 2>&1 &"
+            cmd = f"(timeout 60 {self.config.executable} stop {self.container_id} || {self.config.executable} kill {self.container_id}) >/dev/null 2>&1 &"
             subprocess.Popen(cmd, shell=True)
 
     def __del__(self):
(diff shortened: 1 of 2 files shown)


dannyward630:
Rechecked current head `8ef9f599`. The cleanup change preserves Docker's configured `--rm` policy by falling back from `docker stop` to `docker kill`, and the focused regression test is included. `pre-commit.ci - pr` is green; the branch is mergeable with no unresolved review feedback. Remaining step: maintainer review/merge.

PR #870, opened 2026-06-19
Fix(config): make shell command prompts and tool descriptions shell-a…

### LINKED ISSUE #829: Bash tool launches Dash on Ubuntu environments

klieret:
Closes #829

### DIFF
--- a/src/minisweagent/config/mini.yaml
+++ b/src/minisweagent/config/mini.yaml
@@ -4,7 +4,7 @@ agent:
   instance_template: |
     Please solve this issue: {{task}}
 
-    You can execute bash commands and edit files to implement the necessary changes.
+    You can execute shell commands and edit files to implement the necessary changes.
 
     ## Recommended Workflow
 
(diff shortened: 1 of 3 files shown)


Chesars:
Hi @klieret. I opened [#890](https://github.com/SWE-agent/mini-swe-agent/pull/890) as a draft on top of this branch. It renames the model-facing tool from `bash` to `shell` in the tool-call schema. 
Would this make sense as a follow-up here? Happy to close/re-scope it if not.

PR #871, opened 2026-06-21
Make local command failures explicit in observations

katherinelazar:
## Summary

This makes failed `LocalEnvironment` shell commands more explicit in the observation text.

When a command exits with a positive nonzero return code, the output is prefixed with:

```text
COMMAND FAILED with exit code <code>
```

The original stdout/stderr output is preserved after the marker.

## Motivation

Failed shell commands can be easy for agents to miss when the failure is only visible through the return code or buried in command output. This makes real process exits more obvious without changing agent control flow.

## Scope

This is intentionally scoped to `LocalEnvironment`. Other environment backends are unchanged. A follow-up could move this into shared observation formatting if maintainers want consistent behavior across all environments.

## Testing

```bash
PATH="$PWD/.venv/bin:$PATH" .venv/bin/python -m pytest tests/environments/test_local.py
PATH="$PWD/.venv/bin:$PATH" .venv/bin/python -m pytest tests
```

Results:

```text
20 passed
531 passed, 76 skipped, 1 warning
```

### DIFF
--- a/src/minisweagent/environments/local.py
+++ b/src/minisweagent/environments/local.py
@@ -39,6 +39,7 @@ class LocalEnvironment:
                 "exception_info": f"An error occurred while executing the command: {e}",
                 "extra": {"exception_type": type(e).__name__, "exception": str(e)},
             }
+        _mark_failed_command(output)
         self._check_finished(output)
         return output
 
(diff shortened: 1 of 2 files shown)


klieret:
Hi thanks, but I think the exit code is clear enough for most LMs.

PR #878, opened 2026-06-26
fix(models): bound provider-facing observation output

### LINKED ISSUE #873: mini-swe-agent forwards large shell output back to the provider
## Describe the bug

**Summary**

A provider-requested shell command that writes a large stdout payload causes mini-swe-agent to forward the large result back into the provider loop instead of bounding or truncating it.

The standalone reproducer installs public `mini-swe-agent==2.4.2`, starts a local OpenAI-compatible mock provider, asks mini-swe-agent to run a shell command that prints 100000 bytes, and verifies that the resulting provider traffic contains at least 8192 bytes of tool-result output. A local public-package run observed 30441 bytes of tool-result payload.

**Affected target**

- Target: `mini-swe-agent`
- Affected version checked with the public reproducer: `2.4.2`
- Current public latest checked on 2026-06-22: `2.4.2`
- Environment: Linux x86_64 Docker, public package install
- Provider/model: OpenAI-compatible local mock provider, `gpt-4o`

**Steps to reproduce**

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It installs the affected CLI version from the public package registry, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/1620a2feef749941e2f86ab2499c8ca4

The Gist contains `mini-swe-agent-huge-stdout-resource-bound.reproduce.py`. Download and run:

```bash
gh gist clone 1620a2feef749941e2f86ab2499c8ca4 mini-swe-agent-huge-stdout-resource-bound-reproducer
cd mini-swe-agent-huge-stdout-resource-bound-reproducer
python3 mini-swe-agent-huge-stdout-resource-bound.reproduce.py
```

To reuse an already-built local image:

```bash
python3 mini-swe-agent-huge-stdout-resource-bound.reproduce.py --skip-build
```

Expected successful reproduction output includes:

```text
provider_requests_at_least: observed=4 expected>=2
tool_result_bytes_at_least: observed=30441 expected>=8192
RUN_REPRODUCED
REPRODUCED
```

**Actual behavior**

mini-swe-agent forwards the large shell stdout payload back to the provider as tool-result content. In the local standalone run, the script observed 30441 bytes of tool-result payload for the large-output command.

**Expected behavior**

mini-swe-agent should bound, truncate, or otherwise protect provider-facing tool-result output so a single command cannot amplify a large stdout payload into the model interaction.

**Public-upstream status**

This report targets stock public `mini-swe-agent==2.4.2`, which is the current PyPI latest checked on 2026-06-22. No upstream issue has been filed from this draft yet. This is separate from the repeated-tool report because it validates provider-facing output bounding for a single large shell result.

snvtac:
## Summary

- Add a shared observation utility that bounds provider-facing command output while preserving the original `extra.raw_output` for local trajectory/debug use.
- Apply the bound before rendering observations for tool calls, Responses API tool calls, and text-based observations.
- Add regression coverage for large stdout, retry-amplified provider payloads, the real `mini.yaml` observation template, and provider message preparation stripping `extra.raw_output`.

Fixes #873

## Validation

- `uv run pytest tests/models/test_observation_truncation.py`
- `uv run pytest tests/models -n auto` -> 140 passed, 1 warning
- `uv run pytest -n auto` -> 536 passed, 76 skipped, 1 warning
- `uv run ruff check src tests`
- `uv run ruff format --check src tests`
- `uv run pylint src/minisweagent --errors-only`
- `git diff --check`

## Notes

Validation used a repo-local `.venv` and `.uv-cache`; both were removed after verification along with ignored test/lint caches.

### DIFF
--- a/src/minisweagent/models/utils/actions_text.py
+++ b/src/minisweagent/models/utils/actions_text.py
@@ -9,6 +9,7 @@ import time
 from jinja2 import StrictUndefined, Template
 
 from minisweagent.exceptions import FormatError
+from minisweagent.models.utils.observations import bounded_observation_output
 from minisweagent.models.utils.openai_multimodal import expand_multimodal_content
 
 
@@ -50,8 +51,9 @@ def format_observation_messages(
     """Format execution outputs into user observation messages."""
     results = []
     for output in outputs:
+        bounded_output = bounded_observation_output(output)
         content = Template(observation_template, undefined=StrictUndefined).render(
-            output=output, **(template_vars or {})
+            output=bounded_output, **(template_vars or {})
         )
         msg: dict = {
             "role": "user",
--- a/src/minisweagent/models/utils/actions_toolcall.py
+++ b/src/minisweagent/models/utils/actions_toolcall.py
@@ -6,6 +6,7 @@ import time
 from jinja2 import StrictUndefined, Template
 
 from minisweagent.exceptions import FormatError
+from minisweagent.models.utils.observations import bounded_observation_output
 from minisweagent.models.utils.openai_multimodal import expand_multimodal_content
 
 BASH_TOOL = {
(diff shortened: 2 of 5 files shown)


klieret:
But we can already truncate observations with the jinja templates, why is this needed?

klieret:
see the `.yaml` configs

PR #880, opened 2026-06-27
fix(run): pass the startup_command as a dict to env.execute() function.

Ethan-Kuang-6666:
## Summary

The get_sb_environment() function in swebench.py is used whenever an environment is needed while running benchmarks. If there exist environment startup commands, they will be passed to the env.execute() function to get executed. However, the env.execute() function expects the commands to be passed as a dict, and in current version the startup_command is passed as a string. An AttributeError will be thrown.

## The Fix

startup_command is not wrapped inside a dict as {"command": startup_command} and passed to env.execute().

## Test

Unit test test_get_sb_environment_runs_startup_command_as_dict() was added to test this exact fix.


klieret:
oh! very good catch! Tysm!

klieret:
Congrats to your first contrib, now part of 2.4.3 ❤️

PR #881, opened 2026-06-29
fix(env): atexit cleanup for DockerEnvironment

vkandbro:
## Problem

The `DockerEnvironment.cleanup()` method was only called via `__del__`, which is not guaranteed to execute on abnormal exit (Ctrl+C, interpreter crash). This could leave Docker containers running as zombie processes.

## Solution

Register `atexit.register(self.cleanup)` in `__init__`, which guarantees execution on normal Python exit. Add a `_cleaned` flag to make `cleanup()` idempotent for the case where both `atexit` and `__del__` fire.

## Changes

- `src/minisweagent/environments/docker.py`: Add `atexit.register`, add `_cleaned` idempotency flag
- `tests/environments/test_docker_cleanup.py`: New tests for cleanup idempotency

## Testing

- `test_cleanup_idempotent`: Verifies multiple `cleanup()` calls do not raise
- `test_cleaned_flag`: Verifies `_cleaned` is set after cleanup

## Notes for Reviewer

- No API changes — all existing usage patterns remain compatible
- `atexit` is stdlib, no new dependencies
- `_cleaned` flag protects against the atexit + `__del__` double-call edge case

### DIFF
--- a/src/minisweagent/environments/docker.py
+++ b/src/minisweagent/environments/docker.py
@@ -1,7 +1,9 @@
+import atexit
 import logging
 import os
 import platform
 import shlex
+import signal
 import subprocess
 import uuid
 from typing import Any
@@ -56,6 +58,8 @@ class DockerEnvironment:
         self.logger = logger or logging.getLogger("minisweagent.environment")
         self.container_id: str | None = None
         self.config = config_class(**kwargs)
+        self._cleaned = False
+        atexit.register(self.cleanup)
         self._start_container()
 
     def get_template_vars(self, **kwargs) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


src/minisweagent/environments/docker.py
klieret, line 165: could we remove this then?

klieret (commented):
Oh, you're right, this is better! But should we remove the `__del__` then?

klieret:
OK so the bigger problem here is that right now basically we rely on python garbage collection to clean up docker containers. Once the env is GCd `__del__` is called which cleans up the container. However, with the `atexit` call, we'd now keep a reference to this object, so GC will never trigger anymore. Meaning that the automatic cleanup is no longer active.

PR #883, opened 2026-06-29
Reject empty config override keys

CodingFeng101:
## What changed

This rejects empty keys in command-line config override specs such as `=value`, `model..name=value`, or `model.=value`.

## Why

`-c key=value` overrides are merged into the agent config. Empty key segments are almost always typos, but they were previously accepted and produced dictionaries with `""` keys. That can silently pollute the merged config and make the real mistake harder to diagnose.

Valid overrides and JSON value parsing are unchanged.

## Validation

- `PYTHONPATH=src python -m pytest tests\config\test_init.py -q`
- `python -m ruff check src\minisweagent\config\__init__.py tests\config\test_init.py`
- `python -m ruff format --check src\minisweagent\config\__init__.py tests\config\test_init.py`
- `git diff --check`

### DIFF
--- a/src/minisweagent/config/__init__.py
+++ b/src/minisweagent/config/__init__.py
@@ -42,6 +42,8 @@ def _key_value_spec_to_nested_dict(config_spec: str) -> dict:
     except json.JSONDecodeError:
         pass
     keys = key.split(".")
+    if any(k == "" for k in keys):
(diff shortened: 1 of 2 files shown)


klieret (approved):
You're right, good point

klieret:
Congrats to your first contrib, now part of 2.4.3 ❤️

PR #891, opened 2026-07-05
fix(env): handle ProcessLookupError in local env timeout kill

C0d3N1nja97342:
## Problem

When a command times out in the local environment, `_run` calls `os.killpg(process.pid, signal.SIGKILL)` to kill the process group. If the process group has already exited between `communicate(timeout=...)` raising `TimeoutExpired` and the `killpg` call (a race condition), `killpg` raises `ProcessLookupError`.

This causes two problems:
1. **Resource leak + lost output:** the subsequent `process.communicate()` (which drains the stdout pipe) is skipped, leaking the file descriptor and losing the partial output that was already buffered.
2. **Wrong exception type:** the caller receives `ProcessLookupError` instead of `TimeoutExpired`. Since `ProcessLookupError` has no `output` attribute, the partial output is irrecoverably lost (`getattr(e, "output", None)` returns `None`).

## Solution

Wrap the `killpg`/`kill` call in `try/except ProcessLookupError: pass` so that `communicate()` always runs and the partial output is captured in the `TimeoutExpired` exception.

## Changes

- `src/minisweagent/environments/local.py`: wrap the kill call in `try/except ProcessLookupError`.
- `tests/environments/test_local.py`: add `test_local_environment_timeout_killpg_race_captures_output` (POSIX-only) that mocks `os.killpg` to raise `ProcessLookupError` and asserts partial output is still captured.

## Testing

- New test passes on Linux: `pytest tests/environments/test_local.py -k "timeout or killpg_race"` → **4 passed** (including the new test + existing timeout tests, no regression).
- `ruff check` clean on changed files.
- The test is POSIX-only (`@pytest.mark.skipif(os.name == "nt")`) since `os.killpg` is POSIX-specific.

### DIFF
--- a/src/minisweagent/environments/local.py
+++ b/src/minisweagent/environments/local.py
@@ -86,7 +86,10 @@ def _run(command: str, cwd: str, env: dict[str, str], timeout: int) -> subproces
     try:
         stdout, _ = process.communicate(timeout=timeout)
     except subprocess.TimeoutExpired:
-        os.killpg(process.pid, signal.SIGKILL) if os.name == "posix" else process.kill()
+        try:
+            os.killpg(process.pid, signal.SIGKILL) if os.name == "posix" else process.kill()
+        except ProcessLookupError:
+            pass  # Process group already exited between timeout and kill.
         stdout, _ = process.communicate()
         raise subprocess.TimeoutExpired(command, timeout, output=stdout)
     return subprocess.CompletedProcess(command, process.returncode, stdout=stdout)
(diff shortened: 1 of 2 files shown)


C0d3N1nja97342:
Related issue: #893

klieret:
Is this an error you actually encountered in the wild?

C0d3N1nja97342:
Closing this — the race is hard to trigger in practice and the impact is minimal, so the issue was marked not-planned. Thanks for the feedback!

PR #892, opened 2026-07-05
fix(config): read config files as UTF-8 regardless of locale

EnesYilmazcode:
## Summary

`get_config_from_spec` loaded config files with `path.read_text()`, which decodes using the ambient locale codec. The shipped `programbench.yaml` contains a UTF-8 em dash, so under a C/ascii locale (common in CI and minimal containers) the read raises `UnicodeDecodeError` and the run crashes at startup; on a Windows cp1252 console it silently mis-decodes the prompt. Read configs as UTF-8 so loading no longer depends on the ambient locale.

## Test plan

- Added `test_loads_utf8_yaml_regardless_of_locale`: runs a subprocess under `LC_ALL=C` / `PYTHONUTF8=0`, writes a UTF-8 config containing an em dash, and asserts it parses. Fails on the old code (`UnicodeDecodeError`), passes with the fix.
- `uv run pytest tests/config/test_init.py` -> 38 passed.

### DIFF
--- a/src/minisweagent/config/__init__.py
+++ b/src/minisweagent/config/__init__.py
@@ -58,7 +58,7 @@ def get_config_from_spec(config_spec: str | Path) -> dict:
     if isinstance(config_spec, str) and "=" in config_spec:
         return _key_value_spec_to_nested_dict(config_spec)
     path = get_config_path(config_spec)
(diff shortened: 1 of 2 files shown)


klieret (approved):

klieret:
thx

PR #900, opened 2026-07-07
fix(config): default.yaml format_error_template describes tool-call runtime it wires up

### LINKED ISSUE #899: The default model class uses the wrong `format_error_template`
### Describe the bug

[default.yaml](https://github.com/SWE-agent/mini-swe-agent/blob/e187bcb2ff5825d85761a6f9c1f98c9fa6cfbc79/src/minisweagent/config/default.yaml#L144) uses the template same as that in [mini_textbased.yaml](https://github.com/SWE-agent/mini-swe-agent/blob/e187bcb2ff5825d85761a6f9c1f98c9fa6cfbc79/src/minisweagent/config/mini_textbased.yaml#L145) instead of [mini.yaml](https://github.com/SWE-agent/mini-swe-agent/blob/e187bcb2ff5825d85761a6f9c1f98c9fa6cfbc79/src/minisweagent/config/mini.yaml#L129), which is wrong.

This sometimes steers model into an non-recoverable path to eventual `RepeatedFormatError` when model's first attempt is in a wrong format if mini is using the default model class

yyl:
`default.yaml`'s `format_error_template` describes text-fence recovery, but the same file wires up `DefaultAgent` + `LitellmModel` — a tool-call runtime. When a model returns `content` without `tool_calls`, `parse_toolcall_actions` (`src/minisweagent/models/utils/actions_toolcall.py:40`) raises `FormatError` with a recovery hint pointing at ```` ```mswea_bash_command ```` fences the parser doesn't check.

Fix https://github.com/SWE-agent/mini-swe-agent/issues/899

## Test plan

- [x] `diff <(sed -n '144,164p' src/minisweagent/config/default.yaml) <(sed -n '129,149p' src/minisweagent/config/mini.yaml)` returns empty.
- [x] `tests/models/test_truncation_finish_reason.py` — 13 pass; already covers the `has_tool_calls` branch.

### DIFF
--- a/src/minisweagent/config/default.yaml
+++ b/src/minisweagent/config/default.yaml
@@ -142,30 +142,23 @@ model:
   model_kwargs:
     drop_params: true
   format_error_template: |
-    {% if finish_reason is defined and finish_reason in ["length", "tool_calls"] -%}
-    Your previous response reached the output token limit (finish_reason={{ finish_reason }}) before you produced a complete action, so it was cut off. Respond more concisely and provide exactly one action in the required format. If you need to think more, do so briefly.
+    {% if finish_reason is defined and (finish_reason == "length" or (finish_reason == "tool_calls" and not has_tool_calls)) -%}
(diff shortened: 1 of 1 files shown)


klieret:
oh I guess default.yaml is still built around non-toolcalls. But it's also not really used by anything but demos, so might actually just remove the whole config

PR #903, opened 2026-07-10
enh(run): defer prompt_toolkit import until prompting

dexhunter:
## Summary

- defer the private multiline prompt import until the existing no-task branch needs it
- add a fresh-interpreter regression test for the import boundary
- preserve interactive prompting by patching the helper at its defining module

This follows a review suggestion from [PR #749](https://github.com/SWE-agent/mini-swe-agent/pull/749#discussion_r2830036314).

## Performance

Measured with Python 3.11 in the same isolated environment, using 31 fresh interpreters after 5 warmups:

| Revision | Median `minisweagent.run.mini` import |
| --- | ---: |
| Current main (`e187bcb2`) | 118.993 ms |
| This change | 54.976 ms |

That is a 53.80% reduction (2.16x faster) for fresh module import.

This benchmark measures fresh module import only. CLI help is a correctness gate, not a separately timed metric. It does not measure normal default agent execution, and `prompt_toolkit` remains a required dependency.

## Validation

- new subprocess regression test fails on pristine main and passes with this change
- focused tests: 133 passed
- full suite: 568 passed, 51 skipped
- `uvx ruff check --fix .`
- `uvx ruff format --check .`
- `pylint minisweagent/ --errors-only`
- `git diff --check`

## Autoresearch

Supplementary autoresearch: [20-step public dashboard](https://dashboard.weco.ai/share/xdYZU-szc1VeENC-ALLiGDytSoC-bo7Q).

### DIFF
--- a/src/minisweagent/run/mini.py
+++ b/src/minisweagent/run/mini.py
@@ -12,7 +12,6 @@ from rich.console import Console
 
 from minisweagent import global_config_dir
 from minisweagent.agents import get_agent
-from minisweagent.agents.utils.prompt_user import _multiline_prompt
 from minisweagent.config import builtin_config_dir, get_config_from_spec
 from minisweagent.environments import get_environment
 from minisweagent.models import get_model
@@ -92,6 +91,8 @@ def main(
     config = recursive_merge(*configs)
 
     if (run_task := config.get("run", {}).get("task", UNSET)) is UNSET:
+        from minisweagent.agents.utils.prompt_user import _multiline_prompt
+
         console.print("[bold yellow]What do you want to do?")
         run_task = _multiline_prompt()
         console.print("[bold green]Got that, thanks![/bold green]")
--- a/tests/run/test_cli_integration.py
+++ b/tests/run/test_cli_integration.py
@@ -15,6 +15,22 @@ def strip_ansi_codes(text: str) -> str:
     return ansi_escape.sub("", text)
 
 
+def test_import_mini_does_not_load_prompt_toolkit():
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-c",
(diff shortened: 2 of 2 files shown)


klieret (approved):

dexhunter:
The only red check is the docs dependency-install step, before repository code runs. The job selected Python 3.14; resolving the current full extra pulled litellm 1.92.0, whose pyo3-ffi 0.23.5 build supports at most Python 3.13 and exits during wheel construction. This PR changes only the prompt_toolkit import location and does not modify dependencies or workflows. The other six checks pass, modified lines are covered, and the PR is approved. The docs job should pass once that environment uses Python 3.13 or the upstream PyO3 constraint is updated.

dexhunter:
Rebased onto current main at `5cb8970`. The full current check set is now green, including the previously failing docs deploy job; the focused lazy-import tests and approval remain in place. This is merge-ready from my side.

dexhunter:
Re-verified against today's `main` (`a83fcae`), four commits past the approved head:

- the change cherry-picks onto `main` with no conflict, and the top-level `_multiline_prompt` import is still at `mini.py:15`, so it still applies exactly as written;
- the regression test still fails against unmodified `main` and passes with the change;
- re-measured fresh-interpreter `import minisweagent.run.mini` on Python 3.11, 31 runs after 5 warmups with the two arms interleaved: 204.6 ms -> 104.2 ms median. The absolute numbers are higher than in the description because this box is busier today; net of a 15.1 ms bare-interpreter floor that is 189.5 ms -> 89.1 ms, about 2.1x, in line with the original measurement.

Nothing outstanding from my side. Happy to rebase onto current `main`, or to split the regression test out on its own, if either would make it easier.

PR #904, opened 2026-07-11
fix(config): remove default.yaml, which paired text prompts with the tool-call model

### LINKED ISSUE #899: The default model class uses the wrong `format_error_template`

EnesYilmazcode:
## Summary

`default.yaml` told the model to answer in `mswea_bash_command` text blocks and repeated that guidance in its `format_error_template`, but `hello_world.py` paired it with the default model class, which only parses native tool calls. A model that followed the prompts could only loop through format errors into `RepeatedFormatError`.

Rather than fixing the template, this deletes the config, as suggested in [#899](https://github.com/SWE-agent/mini-swe-agent/issues/899) ("might actually just get rid of `default.yaml`? I guess it's only used in `hello_world.py` right now") and on [#900](https://github.com/SWE-agent/mini-swe-agent/pull/900) ("it's also not really used by anything but demos, so might actually just remove the whole config").

Changes:

- `hello_world.py` (the only runtime consumer) now loads `mini.yaml`
- the four test files that used `default.yaml` as their text-format agent config (`tests/agents/test_default.py`, `tests/agents/test_interactive.py`, `tests/run/test_cli_integration.py`, `tests/run/test_save.py`) now load `mini_textbased.yaml`, which has the same text templates
- dropped the `default.yaml` line from the config README

No references to `default.yaml` remain after this.

## Test

The four updated test files give identical results on this branch and on `main`. `DefaultAgent` instantiates cleanly with `mini.yaml`'s agent config, so `hello_world.py` still works.

Fixes #899


MohammedAlkindi (commented):
Ran this on Windows 11 (Python 3.13), since CI here is Linux-only and this PR deletes a config four test files load. Full suite both sides, `tests/environments/extra` excluded (needs `contree_sdk`/modal SDKs I don't have).

**Base** (current `main`, `a83fcae8`):

```
22 failed, 349 passed, 60 skipped, 156 errors   (587 collected)
```

**With this PR** (merged onto the same `main` locally — it is 5 commits behind but auto-merges cleanly, `tests/agents/test_default.py` being the only overlap):

```
22 failed, 348 passed, 60 skipped, 156 errors   (586 collected)
```

I diffed the two `FAILED`/`ERROR` lists: **byte-identical, 178 lines each**. The 22 failures and 156 errors are pre-existing Windows breakage on `main` (mostly prompt_toolkit pipe-input setup errors in `test_interactive.py`), untouched either way.

The −1 pass / −1 collected is expected and benign: `tests/config/test_init.py` builds `_ALL_BUILTIN_CONFIGS` from `rglob("*.yaml")`, so deleting `default.yaml` removes exactly that parametrized case — i.e., the per-config validation now runs over the remaining configs and passes.

Also verified the PR body's claim directly: `grep -rn 'default\.yaml' src/ tests/ docs/` on the branch returns nothing (my only hit was a local `egg-info/SOURCES.txt` build artifact from the editable install). `hello_world.py` now loads `config/mini.yaml`, which pairs the text prompts with the text parser rather than the tool-call model — the mismatch #899 reported.

Two disclosures on method: (1) the suite cannot even be *collected* on headless Windows without the lazy `PromptSession` construction from #934 (`prompt_user.py` raises `NoConsoleScreenBufferError` at import), so I applied that two-line change identically to both sides; the comparison isolates this PR. (2) I did not run `tests/environments/extra` on either side, as noted.

From a Windows run's perspective this does what it says, removes the #899 trap, and introduces no regression.

MohammedAlkindi:
Correcting my own numbers above — the conclusion stands, but the absolute figures were distorted by my runner and I'd rather fix that here than leave them quotable.

My earlier runs invoked pytest with absolute paths from *outside* the repo root. Several fixtures resolve config paths relative to the cwd (e.g. `Path("src/minisweagent/config/mini_textbased.yaml")` in `tests/agents/test_interactive.py:108`), so a large batch of tests errored at setup with `FileNotFoundError` before doing any work, and I over-attributed the wreckage to prompt_toolkit. Re-run correctly, from the repo root, same machine, same venv:

- base (`main` + the #934 import fix): `144 failed, 383 passed, 60 skipped`
- this PR on the same base: `144 failed, 382 passed, 60 skipped`

The two `FAILED` lists are **identical** (144 lines both sides, diffed empty), and the −1 pass is still the `rglob("*.yaml")` parametrization losing the deleted file. So the review's actual claim — this PR introduces no regression on Windows, and the `default.yaml` removal is clean — is unchanged, now on cleaner methodology.

(The real Windows state of `main` is therefore "144 failures from the repo root", dominated by `tests/agents/test_interactive.py` desyncing from the scripted test model — a separate story from this PR either way, and one I haven't diagnosed.)

PR #915, opened 2026-07-21
fix(agents): count billed calls that fail to parse against cost_limit

### LINKED ISSUE #914: Format errors are billed but not counted, so the agent runs past cost_limit
### Describe the bug

`model.query()` charges for the API call before it parses it. In [`litellm_model.py:85-97`](https://github.com/SWE-agent/mini-swe-agent/blob/main/src/minisweagent/models/litellm_model.py#L85-L97) the order is: `_query()` makes the call, `_calculate_cost(response)` computes the cost, `GLOBAL_MODEL_STATS.add(...)` records it, and then `_parse_actions(response)` can raise `FormatError`, which is re-raised. So `query()` never returns.

In the agent:

```python
self.n_calls += 1
message = self.model.query(self.messages)                 # default.py:147, raises here
self.cost += message.get("extra", {}).get("cost", 0.0)    # default.py:148, never runs
```

`run()` catches the `FormatError` and keeps looping, so the next turn is billed too. Since `default.py:130` checks `self.cost` against `config.cost_limit`, a turn that fails to parse is invisible to the budget.

The tell is that `GLOBAL_MODEL_STATS.cost` ends up with the right total while `agent.cost` does not. The number is known, it just never reaches the agent.

### Impact

With `cost_limit=2.5` and $1.00 per call, an agent whose turns all fail to parse should stop after 3 calls. Instead it never stops on cost, and runs until some other limit catches it. `self.cost` is also what is written as `instance_cost` in the trajectory, so a run under-reports its spend by whatever the format-error turns cost.

This is reachable in normal operation rather than a corner case: `max_consecutive_format_errors` and #863 exist because models do repeatedly emit unparseable turns.

### Affected

All seven `query()` implementations bill before parsing: `litellm_model`, `litellm_response_model`, `openrouter_model`, `openrouter_response_model`, `portkey_model`, `portkey_response_model`, `requesty_model`.

### Suggested fix

`litellm_model.py:87` already states the contract that "all model.query() implementations must persist the response on FormatError". Extending that to cover the cost, and reading it in the `except FormatError` handler `run()` already has, fixes this without adding new control flow. Happy to open a PR.

EnesYilmazcode:
## Summary

A model call is billed before its response is parsed. In `litellm_model.py` `query()` runs `_calculate_cost()` and `GLOBAL_MODEL_STATS.add(...)`, then `_parse_actions()` can raise `FormatError` and re-raise, so `query()` never returns. The agent's `self.cost += ...` on the next line never runs, `run()` catches the error and loops, and the budget check at `default.py:130` never sees the spend. A run keeps paying past its `cost_limit`, and `instance_cost` in the trajectory comes out low.

The cost was never lost, only stranded: `GLOBAL_MODEL_STATS` had the right total the whole time while `agent.cost` did not.

`litellm_model.py:87` already says "all model.query() implementations must persist the response on FormatError". This extends that contract to the cost and reads it from the `except FormatError` handler `run()` already has, so no new control flow is introduced.

Fixes #914.

## Changes

- Each of the seven `query()` implementations adds `e.messages[0]["extra"].update(cost_output)` inside the `except FormatError` block that already persists the response.
- `DefaultAgent.run()` adds the persisted cost in its existing handler.
- The contract note in `litellm_model.py` now reads "response and cost".

Production change is 11 lines. The rest is tests.

## Test plan

- [x] New `test_format_errors_count_against_cost_limit` in `tests/agents/test_default.py`: an agent whose turns all fail to parse stops at `cost_limit` after 3 calls instead of running to `step_limit`. Removing the `default.py` change makes it fail with `8 != 3`.
- [x] The seven existing per-model tests in `tests/models/test_format_error_response_persistence.py` now also assert the persisted cost. Verified by mutation: deleting the new line from any one of the seven fails exactly one test. Before this, deleting it from `litellm_response_model.py` failed nothing.
- [x] Full suite 333 passed vs 332 on `main`, with the same pre-existing failures on both (Windows `prompt_toolkit`, docker and network tests).
- [x] `ruff check` and `ruff format --check` clean.

## Notes

Scope is deliberately narrow. `FormatError` is the only path that bills and then keeps looping, so it is the only one that leaks. The other handlers in `run()` are terminal.

This does not overlap #910, which resets `self.cost` between `run()` invocations. That is the same field for a different reason, and the two changes touch different parts of `run()`.

One unrelated thing I noticed: `test_format_error_counter_resets_on_success` currently fails on a clean `main` with an `IndexError` from an exhausted `outputs` list. Not touched here, but worth a look.

### DIFF
--- a/src/minisweagent/agents/default.py
+++ b/src/minisweagent/agents/default.py
@@ -98,6 +98,8 @@ class DefaultAgent:
                 self.step()
                 self.n_consecutive_format_errors = 0  # reset on any clean step
             except FormatError as e:
+                # The call was billed before parsing failed, so query() never got to charge it.
+                self.cost += e.messages[0].get("extra", {}).get("cost", 0.0)
                 self.n_consecutive_format_errors += 1
                 if 0 < self.config.max_consecutive_format_errors <= self.n_consecutive_format_errors:
                     self.add_messages(
--- a/src/minisweagent/models/litellm_model.py
+++ b/src/minisweagent/models/litellm_model.py
@@ -84,10 +84,11 @@ class LitellmModel:
                 response = self._query(self._prepare_messages_for_api(messages), **kwargs)
         cost_output = self._calculate_cost(response)
         GLOBAL_MODEL_STATS.add(cost_output["cost"])
-        # Note: all model.query() implementations must persist the response on FormatError.
+        # Note: all model.query() implementations must persist the response and cost on FormatError.
         try:
             actions = self._parse_actions(response)
         except FormatError as e:
+            e.messages[0]["extra"].update(cost_output)
             try:
                 e.messages[0]["extra"]["response"] = response.model_dump(mode="json")
             except Exception:
--- a/src/minisweagent/models/litellm_response_model.py
+++ b/src/minisweagent/models/litellm_response_model.py
@@ -58,6 +58,7 @@ class LitellmResponseModel(LitellmModel):
         try:
             actions = self._parse_actions(response)
         except FormatError as e:
+            e.messages[0]["extra"].update(cost_output)
             # hasattr guard: litellm.responses() returns a pydantic object, but tests
             # may inject a plain dict; dict(response) is the correct fallback in that case.
             # Inner try: if serialization itself fails, repr() guarantees the key is always set.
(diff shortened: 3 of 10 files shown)


klieret:
wait sorry, I think I pushed a nonsensical refactor here by accident. But the issue you point out is real, let's fix it. Just was hoping that there might be an easier way, but probably there isn't one without large refactors

klieret:
Part of v2.4.6 now, thanks again!

PR #919, opened 2026-07-29
Add MiniMax-M3 and MiniMax-M2.7 model configuration

octo-patch:
Reason: The SWE-bench XML benchmark config and the model quickstart docs only referenced the older `minimax/minimax-m2`, with no configuration for the newer MiniMax-M3 or MiniMax-M2.7 models.

## Changes

- `src/minisweagent/config/benchmarks/swebench_xml.yaml`: update the OpenRouter model used by the executable benchmark configuration from `minimax/minimax-m2` to the current flagship `minimax/minimax-m3`.
- `docs/models/quickstart.md`: replace the single MiniMax example with two OpenRouter examples so both current models are documented — `minimax/minimax-m3` and `minimax/minimax-m2.7`.

Both continue to use the existing `openrouter` model class and the same `temperature: 0.0` model kwargs already used for MiniMax, so no code paths change.

## Checks

- `python3 -c "import yaml; yaml.safe_load(open('src/minisweagent/config/benchmarks/swebench_xml.yaml'))"` — the benchmark config still parses and resolves `model_name`/`model_class`.

### DIFF
--- a/docs/models/quickstart.md
+++ b/docs/models/quickstart.md
@@ -243,11 +243,21 @@ Here are more examples of how to configure specific models:
             temperature: 0.0
     ```
 
-=== "Minimax (Openrouter)"
+=== "Minimax M3 (Openrouter)"
 
     ```yaml
     model:
-        model_name: "minimax/minimax-m2"
+        model_name: "minimax/minimax-m3"
+        model_class: openrouter
+        model_kwargs:
+            temperature: 0.0
+    ```
+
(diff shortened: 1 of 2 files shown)


klieret:
hi, thx, I just want to avoid too much clutter, I'm ok with adding one of them

octo-patch:
Thanks for the review. I kept the quickstart focused on the current MiniMax-M3 example and removed the obsolete duplicate entry, then pushed a new commit. The documentation diff check passed.

PR #922, opened 2026-08-02
fix(models): stop retrying context window errors

### LINKED ISSUE #921: Permanent context-overflow (400) retried 10 times — full oversized history resent each attempt
## Describe the bug

When the provider returns HTTP 400 (context window exceeded), the OpenRouter and Requesty adapters map it to a generic `OpenRouterAPIError` / `RequestyAPIError` (`src/minisweagent/models/openrouter_model.py:88`, `requesty_model.py:93` — the `else` branch that catches any non-401/non-429 error). This exception is not in `abort_exceptions` (`[OpenRouterAuthenticationError, KeyboardInterrupt]`), so the shared retry factory (`src/minisweagent/models/utils/retry.py:19-25`, `stop_after_attempt(10)` + `retry_if_not_exception_type(abort_exceptions)`) treats it as retryable.

A 400 from context overflow is permanent — retrying with the same history will always fail. The request is retried 10 times total (1 initial + 9 retries) with backoff `[4, 4, 4, 8, 16, 32, 60, 60, 60]` seconds.

## Steps to reproduce

Verified with a **real HTTP 400 response** (local server returning 400, not a mock of `_query`):

```bash
git clone https://github.com/SWE-agent/mini-swe-agent
cd mini-swe-agent && pip install -e .
git checkout e187bcb  # v2.4.5

# 1. Start a fake server that returns HTTP 400
python3 -c "
import http.server, json
class H(http.server.BaseHTTPRequestHandler):
    def do_POST(self):
        b = json.dumps({'error': {'message': 'context_length_exceeded', 'type': 'invalid_request_error', 'code': 'context_length_exceeded'}}).encode()
        self.send_response(400)
        self.send_header('Content-Type', 'application/json')
        self.send_header('Content-Length', str(len(b)))
        self.end_headers()
        self.wfile.write(b)
    def log_message(self, *a): pass
http.server.HTTPServer(('127.0.0.1', 18399), H).serve_forever()
" &

# 2. Point OpenRouterModel at it and send an oversized request
OPENROUTER_API_KEY=fake python3 -c "
import sys, time
sys.path.insert(0, 'src')
# Patch time.sleep to skip the 248s of backoff waiting
time.sleep = lambda s: None
from minisweagent.models.openrouter_model import OpenRouterModel, OpenRouterAPIError
model = OpenRouterModel(model_name='test', completion_kwargs={'api_base': 'http://127.0.0.1:18399/v1'})
model._api_url = 'http://127.0.0.1:18399/v1/chat/completions'
try:
    model.query([{'role': 'user', 'content': 'x' * 999999}])
except OpenRouterAPIError:
    pass
"
```

The HTTP call, exception mapping (`HTTPError` → `OpenRouterAPIError` at line 88), and retry logic are all unmodified production code. Only `time.sleep` is patched to skip backoff delays.

## Expected behavior

A permanent 400 (context overflow) should abort immediately, like 401 authentication errors do (`OpenRouterAuthenticationError` is in `abort_exceptions`).

## Actual behavior

**10 real HTTP round-trips** to the 400 server, then the error surfaces. Total backoff wait: 248 seconds.

The `MSWEA_MODEL_RETRY_STOP_AFTER_ATTEMPT` env var can override the default of 10, but the default shipped behavior retries permanent failures.

## Suggested fix

Add `context_length_exceeded` (or the specific 400 status) to `abort_exceptions` in `openrouter_model.py` and `requesty_model.py` — not the generic `APIError`, which would also abort legitimate 500/network retries.

Tested with mini-swe-agent v2.4.5 (commit `e187bcb`).

swayamg20:
## Summary
- Classify only HTTP 400 context-window overflows using canonical provider error fields and a conservative message fallback.
- Raise provider-specific API-error subclasses across the OpenRouter tool-call, text-based, and Responses adapters, plus Requesty, so permanent failures abort immediately.
- Preserve existing retries for unrelated 400 responses, transient 5xx responses, and network errors.

Closes #921

## Testing
- Real local HTTP-server tests verify context overflows make one request while other API errors retain the configured retry count.
- `pytest -q tests/models/test_context_window_errors.py tests/models/test_openrouter_textbased_model.py` (16 passed)
- `pytest -q -n auto` (552 passed, 76 skipped)
- Ruff check and format validation on all changed files
- All applicable Python pre-commit hooks on all changed files

### DIFF
--- a/src/minisweagent/models/openrouter_model.py
+++ b/src/minisweagent/models/openrouter_model.py
@@ -15,6 +15,7 @@ from minisweagent.models.utils.actions_toolcall import (
     parse_toolcall_actions,
 )
 from minisweagent.models.utils.anthropic_utils import _reorder_anthropic_thinking_blocks
+from minisweagent.models.utils.api_errors import is_context_window_error
 from minisweagent.models.utils.cache_control import set_cache_control
 from minisweagent.models.utils.openai_multimodal import expand_multimodal_content
 from minisweagent.models.utils.retry import retry
@@ -44,6 +45,10 @@ class OpenRouterAPIError(Exception):
     """Custom exception for OpenRouter API errors."""
 
 
+class OpenRouterContextWindowError(OpenRouterAPIError):
+    """OpenRouter rejected a request that exceeds the model context window."""
+
+
 class OpenRouterAuthenticationError(Exception):
     """Custom exception for OpenRouter authentication errors."""
 
@@ -53,7 +58,11 @@ class OpenRouterRateLimitError(Exception):
 
 
 class OpenRouterModel:
-    abort_exceptions: list[type[Exception]] = [OpenRouterAuthenticationError, KeyboardInterrupt]
+    abort_exceptions: list[type[Exception]] = [
+        OpenRouterAuthenticationError,
+        OpenRouterContextWindowError,
+        KeyboardInterrupt,
+    ]
 
     def __init__(self, **kwargs):
         self.config = OpenRouterModelConfig(**kwargs)
@@ -84,6 +93,8 @@ class OpenRouterModel:
                 raise OpenRouterAuthenticationError(error_msg) from e
             elif response.status_code == 429:
                 raise OpenRouterRateLimitError("Rate limit exceeded") from e
+            elif is_context_window_error(response):
+                raise OpenRouterContextWindowError(f"HTTP {response.status_code}: {response.text}") from e
             else:
                 raise OpenRouterAPIError(f"HTTP {response.status_code}: {response.text}") from e
         except requests.exceptions.RequestException as e:
--- a/src/minisweagent/models/openrouter_response_model.py
+++ b/src/minisweagent/models/openrouter_response_model.py
@@ -9,6 +9,7 @@ from minisweagent.models import GLOBAL_MODEL_STATS
 from minisweagent.models.openrouter_model import (
     OpenRouterAPIError,
     OpenRouterAuthenticationError,
+    OpenRouterContextWindowError,
     OpenRouterModel,
     OpenRouterModelConfig,
     OpenRouterRateLimitError,
@@ -19,6 +20,7 @@ from minisweagent.models.utils.actions_toolcall_response import (
     format_toolcall_observation_messages,
     parse_toolcall_actions_response,
 )
+from minisweagent.models.utils.api_errors import is_context_window_error
 from minisweagent.models.utils.retry import retry
 
 logger = logging.getLogger("openrouter_response_model")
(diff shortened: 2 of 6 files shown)


EvolveAegis:
Independent corroboration from a black-box PoC (no source patching): against the shipped default `stop_after_attempt(10)`, a `context_length_exceeded` HTTP 400 from the provider goes through up to 10 total attempts (1 initial plus 9 retries) before giving up, while a 401 correctly aborts at attempt 1. We also verified the bug is provider-symmetric: OpenRouter and Requesty share the same `retry.py`, so both exhibit the same retry waste. This PR's `is_context_window_error()` plus a dedicated `ContextWindowError` in `abort_exceptions` moves the 400 from "retry to budget" to "abort at attempt 1", and the added test proving generic 400/500/`token_limit_exceeded` still retry confirms it doesn't over-suppress. One scope note: only `context_length_exceeded`-shaped 400s abort; other permanent 4xx still retry to the budget, worth a follow-up.

swayamg20:
Thanks for independently validating this. Agreed that broader permanent-4xx handling should remain a follow-up, this PR intentionally scopes the non-retry behavior to recognized context-window errors.

MohammedAlkindi (commented):
Two things, one of them actionable.

**First, #923 fixes the same issue.** It says "Fixes #921" as this one says "Closes #921", it touches the same five source files, and it names the same four adapters. Neither references the other. They differ in how they detect the overflow: this PR uses canonical provider error fields plus a message fallback, #923 matches the documented `context_length_exceeded` and `context_window_exceeded` codes. That is a real design choice, broader detection against narrower, and worth deciding deliberately rather than by whichever gets read first.

**Second, and this is the actionable part: the new tests here are flaky when the models suite runs together.** Windows 11, Python 3.13.13.

In isolation the file is solid, three runs in a row:

```
pytest tests/models/test_context_window_errors.py   ->  9 passed  (x3)
```

Run as part of the directory, three runs in a row:

```
pytest tests/models/   ->  3 failed, 145 passed
                       ->  1 failed, 147 passed
                       ->  1 failed, 147 passed
```

The failure count itself varies, so it is nondeterministic rather than an ordering bug with a fixed victim. The one that failed twice was `test_context_window_errors_abort_without_retry[openrouter-responses-error-type]`, and running that single case on its own passes.

For comparison, on the same machine and the same command, `main` is `139 passed` and #923 is `143 passed` with **zero** failures across three runs each.

I think the cause is the real local HTTP servers. `test_context_window_errors.py` has four references to `HTTPServer`/`localhost`/`bind`, and `test_context_length_retry.py` has none. Real sockets under a suite that also exercises other model adapters is the usual way you get a count that changes run to run.

Worth saying plainly that this may be specific to this machine, since I have not seen your CI. But the isolation-versus-suite split is reproducible here, and it is the kind of thing that shows up later as an intermittent CI failure nobody can pin down.

PR #924, opened 2026-08-04
fix(agents): reset task state for each run

### LINKED ISSUE #909: DefaultAgent.run() retains per-run budget and timing state across invocations
### Describe the bug

I found that `DefaultAgent.run()` clears `messages` but retains per-run state such as `n_calls`, `cost`, consecutive format-error count, and the wall-clock start time.

As a result, reusing one agent instance can cause a later task to immediately hit an existing limit. For example, with `step_limit=1`, a second run exits with `LimitsExceeded` before another model query.

Would you be interested in a small PR that resets this task-scoped state at the start of `run()`, with regression coverage? Or is single-use agent behavior intentional / already planned?

daleselaji-dev:
Fixes #909.

## Root cause

`DefaultAgent` initializes its step count, cost, consecutive format-error count, and wall-clock start time only in `__init__()`. Although `run()` clears the message history, it keeps those task-scoped values.

Reusing an agent therefore gives the next task the previous task's consumed budget. A second run can stop with `LimitsExceeded` or `TimeExceeded` before making a model call, and stale format errors can contribute to an unrelated task.

## What changed

- Reset `cost`, `n_calls`, and `n_consecutive_format_errors` at the start of every `run()`.
- Start a fresh wall-clock budget for every `run()`.
- Add a regression test that reuses one agent for two tasks, injects stale cost/error/time state between them, and verifies that the second task receives one independent model call.

This keeps instance configuration, model/environment objects, global model statistics, and per-run trajectory serialization unchanged.

## Validation

```text
python -m pytest +  tests/agents/test_default.py::test_step_limit_enforcement +  tests/agents/test_default.py::test_run_resets_task_scoped_limits -q
6 passed

ruff check src/minisweagent/agents/default.py tests/agents/test_default.py
All checks passed!

ruff format --check src/minisweagent/agents/default.py tests/agents/test_default.py
2 files already formatted
```

The six pytest cases cover the text, tool-call, and Responses API deterministic model fixtures. I did not report the entire `test_default.py` file as green on this Windows host because several existing tests depend on Bash output semantics; the focused limit/reset cases are platform-independent and pass.

## Risk and review notes

The behavioral change is intentionally limited to repeated calls to `run()` on the same agent instance. The primary review point is whether all four values are task-scoped; they already describe per-run limits and are emitted as per-instance trajectory statistics, while global accounting remains in `GLOBAL_MODEL_STATS`.

### DIFF
--- a/src/minisweagent/agents/default.py
+++ b/src/minisweagent/agents/default.py
@@ -87,6 +87,10 @@ class DefaultAgent:
 
     def run(self, task: str = "", **kwargs) -> dict:
         """Run step() until agent is finished. Returns dictionary with exit_status, submission keys."""
+        self.cost = 0.0
+        self.n_calls = 0
+        self.n_consecutive_format_errors = 0
+        self._start_time = time.time()
         self.extra_template_vars |= {"task": task, **kwargs}
         self.messages = []
         self.add_messages(
--- a/tests/agents/test_default.py
+++ b/tests/agents/test_default.py
@@ -170,6 +170,38 @@ def test_step_limit_enforcement(model_factory):
     assert agent.n_calls == 1
 
 
+def test_run_resets_task_scoped_limits(model_factory):
+    """Reusing an agent gives each run independent counters and timing."""
+    factory, config = model_factory
+    model = factory(
+        [
+            ("First command", [{"command": "echo first"}]),
+            ("Second command", [{"command": "echo second"}]),
+        ],
+    )
+    agent = DefaultAgent(
+        model=model,
+        env=LocalEnvironment(),
+        **{
+            **config,
+            "step_limit": 1,
+            "wall_time_limit_seconds": 1,
+        },
+    )
+
+    assert agent.run("First task")["exit_status"] == "LimitsExceeded"
+    agent.cost = 100.0
+    agent.n_consecutive_format_errors = 2
+    agent._start_time = 0
+
+    assert agent.run("Second task")["exit_status"] == "LimitsExceeded"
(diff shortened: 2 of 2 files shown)


MohammedAlkindi (commented):
Flagging that **#910 does the same thing** and has been open since 2026-07-19. Both say "Fixes #909" and neither references the other, so whoever picks this up should probably choose between them rather than review them separately.

The source changes are the same four assignments in `run()`:

```python
self.cost = 0.0
self.n_calls = 0
self.n_consecutive_format_errors = 0
self._start_time = time.time()
```

The only difference is placement. #910 puts them after `self.messages = []`; this PR puts them before `self.extra_template_vars |= {...}`. None of the four is read by either of those two statements, so the ordering has no effect. The tests differ in size, +36 here against +24 on #910.

**#933 is a different matter and should not be lumped in with these.** It says so itself: it fixes `extra_template_vars` accumulating across runs, which neither this PR nor #910 touches, and it already cross-references both. It complements whichever of these lands rather than competing with it.

Verified on Windows 11, Python 3.13.13, comparing deltas rather than absolutes:

```
main    20 passed, 28 failed
#910    22 passed, 28 failed
#924    23 passed, 28 failed
```

The 28 failures are identical on all three and are pre-existing here, not caused by either PR. They are `IndexError: list index out of range` out of `src/minisweagent/models/test_models.py`, the scripted test model running out of queued responses, which is the Windows shell-quoting desync I reported in #937. So on this machine the absolute numbers are not worth much and only the deltas are meaningful.

PR #932, opened 2026-08-14
fix(config): use Notepad as default editor on Windows

### LINKED ISSUE #931: mini-extra config edit fails on Windows when EDITOR is unset
## Description

On Windows, `mini-extra config edit` fails if `EDITOR` is not set.

The current fallback is `nano`:

```python
editor = os.getenv("EDITOR", "nano")
subprocess.run([editor, global_config_file])
```

Since `nano` is usually not available on Windows, this results in:

```text
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

## Reproduction

On Windows:

```powershell
Remove-Item Env:EDITOR -ErrorAction SilentlyContinue
mini-extra config edit
```

## Expected

Use a platform-appropriate fallback such as Notepad on Windows, or show a clearer error telling the user to set `EDITOR`.

Environment: Windows, Python 3.14.6, `EDITOR` unset.

I can submit a small fix with a regression test if this approach makes sense.

ghost:
Summary

use Notepad as the default config editor on Windows
preserve Nano as the default on other platforms
add regression coverage for Windows and POSIX defaults

Fixes #931

### DIFF
--- a/src/minisweagent/run/utilities/config.py
+++ b/src/minisweagent/run/utilities/config.py
@@ -22,6 +22,10 @@ def _reload_config():
     load_dotenv(dotenv_path=global_config_file, override=True)
 
 
+def _default_editor(os_name: str) -> str:
+    return "notepad" if os_name == "nt" else "nano"
+
+
 app = Typer(
     help=__doc__.format(global_config_file=global_config_file),  # type: ignore
     no_args_is_help=True,
@@ -122,7 +126,7 @@ def unset(key: str | None = Argument(None, help="The key to unset")):
 @app.command()
 def edit():
     """Edit the global config file."""
-    editor = os.getenv("EDITOR", "nano")
+    editor = os.getenv("EDITOR", _default_editor(os.name))
     subprocess.run([editor, global_config_file])
     _reload_config()
 
--- a/tests/run/test_extra_config.py
+++ b/tests/run/test_extra_config.py
@@ -3,7 +3,7 @@ from unittest.mock import patch
 
 import pytest
 
-from minisweagent.run.utilities.config import app, configure_if_first_time, edit, set, setup, unset
+from minisweagent.run.utilities.config import _default_editor, app, configure_if_first_time, edit, set, setup, unset
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


MohammedAlkindi:
Tested this branch on a real Windows 11 Home machine (Python 3.13.13, EDITOR unset). At a83fcae, mini-extra config edit crashes exactly as #931 reports: FileNotFoundError WinError 2 from subprocess.run(["nano", ...]), since nano is not on PATH. On this branch, notepad resolves to C:\Windows\System32\notepad.exe and the CLI blocks until Notepad is closed, so _reload_config() still runs after the user saves, verified with a timed probe. pytest tests/run/test_extra_config.py gives 32 passed here vs 30 on main, and the new test_default_editor[nt-notepad] fails on main with an import error. CI cannot see any of this because pytest.yaml runs only on ubuntu-latest. One small follow-up: docs/usage/config.md:44 still documents the fallback as nano only.

MohammedAlkindi (commented):
The fix does what it says, and the test shape is the right one. Windows 11, Python 3.13.13.

`tests/run/test_extra_config.py`: `main` 30 passed, this branch 32 passed. Both green.

The resolution and the spawn, with `EDITOR` unset:

```
main       _default_editor absent, falls back to 'nano'
this PR    _default_editor('nt') -> 'notepad', and the OS launches it
```

**I could not reproduce the original failure on this machine, and the reason is worth knowing.** `nano` is present here at `C:\Program Files\Git\usr\bin\nano.exe`, because **Git for Windows ships it**. So on any Windows box with Git's `usr\bin` on `PATH`, `mini-extra config edit` already works and #931 does not fire. That does not make the bug less real, it narrows who hits it: Windows users without Git for Windows on `PATH`, which is exactly the shape reported.

It also means testing this by spawning a real editor would pass on many Windows machines for the wrong reason. Parameterising `_default_editor(os_name)` and asserting on the mapping is the right call, not a compromise.

One thing the diff deliberately does not do, flagging it only because #931 offered it as an alternative rather than an addition. `subprocess.run([editor, ...])` is unchanged, so an `EDITOR` that is set but not installed still surfaces a bare `FileNotFoundError` traceback. The default is fixed; the explicit-editor path is not. Fine to leave for a separate change, but the issue asked for one or the other and this closes only the first.

Separately, and not a request on this PR since it is identical on both sides: `pytest tests/run` cannot be collected at all on a Windows console here, both on `main` and on this branch:

```
ERROR tests/run/test_local.py - prompt_toolkit.output.win32.NoConsoleScreenBuffer...
Interrupted: 2 errors during collection
```

So `test_extra_config.py` has to be run on its own to see these results. Mentioning it because it means Windows contributors cannot run that directory as a whole, which is a bigger obstacle than the bug this PR fixes.

PR #936, opened 2026-08-15
fix(models): guard missing total_tokens in PortkeyModel cost tracking

### LINKED ISSUE #935: [Bug] PortkeyModel cost tracking crashes when usage.total_tokens is missing (TypeError / negative prompt_tokens)
## Bug Description

`PortkeyModel._calculate_cost()` (src/minisweagent/models/portkey_model.py) crashes or corrupts cost tracking when a provider response is missing `usage.total_tokens`:

1. **If `total_tokens` is `None`**: `total_tokens - prompt_tokens - completion_tokens` raises `TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'`. `_calculate_cost` runs outside the `retry()` loop, so the whole run dies.
2. **If litellm normalizes the missing value to `0`** (its `Usage` default): the consistency check `0 - prompt - completion != 0` fires and rewrites `prompt_tokens = total_tokens - completion_tokens`, producing a **negative** prompt token count (e.g. `prompt_tokens=-5` with prompt=10, completion=5). Costs are then computed against negative tokens.

`prompt_tokens` and `completion_tokens` already have `None` guards; `total_tokens` is the only one left without protection.

## Steps to Reproduce

```python
from types import SimpleNamespace
from litellm import ModelResponse, Usage
from minisweagent.models.portkey_model import PortkeyModel

response = ModelResponse(
    model="gpt-4o",
    usage=Usage(prompt_tokens=10, completion_tokens=5, total_tokens=None),
)
model = PortkeyModel.__new__(PortkeyModel)
model.config = SimpleNamespace(
    litellm_model_name_override="",
    cost_tracking="default",
    model_name="gpt-4o",
)
model._calculate_cost(response)  # TypeError, or prompt_tokens rewritten to -5
```

## Why the obvious fix (zeroing) was rejected before

Related to #781/#779: setting total tokens to 0 is dangerous because cost limits then never trigger and the agent never terminates (maintainer comment on #781). The right fix keeps the real usage: infer `total_tokens` from `prompt_tokens + completion_tokens` with a warning, so the consistency check passes and cost calculation proceeds with actual numbers.

## Expected Behavior

- No `TypeError` when `total_tokens` is `None`
- No negative token counts when `total_tokens` is normalized to `0`
- Cost computed from the real prompt/completion usage

## Environment

- mini-swe-agent 2.4.6
- Windows 11 (bug is provider-data dependent, not platform dependent)

arccat-114:
Fixes #935

## Root cause

`PortkeyModel._calculate_cost()` guards `prompt_tokens` and `completion_tokens` against `None` but not `total_tokens`. Some providers omit `usage.total_tokens`; litellm surfaces it as `None` (→ `TypeError` in `total_tokens - prompt - completion`, crashing the run since `_calculate_cost` runs outside the retry loop) or as `0` (→ the consistency check rewrites `prompt_tokens` to a **negative** number, e.g. -5 for prompt=10/completion=5).

## Fix

When `total_tokens` is `None` or `0`, infer it from `prompt_tokens + completion_tokens` with a warning — instead of zeroing it (zeroing was rejected in #781/#779 because cost limits then never trigger). The consistency check then passes and cost is computed from the real usage.

## Testing

- Added regression test `test_portkey_model_calculate_cost_total_tokens_none` (fails before: prompt_tokens rewritten to -5; passes after)
- `pytest tests/models/`: 140 passed
- ruff: clean

### DIFF
--- a/src/minisweagent/models/portkey_model.py
+++ b/src/minisweagent/models/portkey_model.py
@@ -183,6 +183,16 @@ class PortkeyModel:
                 f"Completion tokens are None for model {self.config.model_name}. Setting to 0. Full response: {response_for_cost_calc.model_dump()}"
             )
             completion_tokens = 0
+        if total_tokens is None or total_tokens == 0:
+            # Some providers omit total_tokens entirely, which litellm
+            # surfaces as None (TypeError below) or 0 (the consistency check
+            # would rewrite prompt_tokens to a negative number).  Inferring
+            # the total from the parts keeps the real usage instead of
+            # silently zeroing it (see #781/#779).
+            logger.warning(
+                f"Total tokens are None or 0 for model {self.config.model_name}. Inferring total from prompt + completion tokens. Full response: {response_for_cost_calc.model_dump()}"
+            )
+            total_tokens = prompt_tokens + completion_tokens
         if total_tokens - prompt_tokens - completion_tokens != 0:
             # This is most likely related to how portkey treats cached tokens: It doesn't count them towards the prompt tokens (?)
             logger.warning(
--- a/tests/models/test_portkey_model.py
+++ b/tests/models/test_portkey_model.py
@@ -191,3 +191,40 @@ def test_portkey_model_cost_validation_error():
 
                 assert "Error calculating cost" in str(exc_info.value)
                 assert "MSWEA_COST_TRACKING='ignore_errors'" in str(exc_info.value)
+
+
+def test_portkey_model_calculate_cost_total_tokens_none():
+    """Regression guard: `usage.total_tokens=None` must not crash cost tracking.
+
+    Portkey sometimes omits ``total_tokens`` for a provider response, and
+    ``_calculate_cost`` would run ``None - prompt - completion`` and raise
+    ``TypeError``.  Unlike setting the value to 0 (rejected in #781/#779
+    because it would make cost limits unreachable), inferring the total from
+    prompt + completion tokens keeps the real usage while avoiding the crash.
+    """
+    from types import SimpleNamespace
+
+    from litellm import ModelResponse, Usage
+
+    response = ModelResponse(
+        model="gpt-4o",
+        usage=Usage(prompt_tokens=10, completion_tokens=5, total_tokens=None),
+    )
+    model = PortkeyModel.__new__(PortkeyModel)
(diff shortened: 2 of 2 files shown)


MohammedAlkindi:
Ran this PR's tests locally (Windows 11, Python 3.13.13, litellm 1.97.0, portkey-ai 2.3.4; `main` @ 25941c8, PR head 7bef506):

**Fail-before** — this PR's test file against unmodified `main` src:

```
FAILED tests/models/test_portkey_model.py::test_portkey_model_calculate_cost_total_tokens_none
1 failed, 7 passed in 41.17s
```

It fails through the *negative-rewrite* path from #935: `_calculate_cost` saw `total_tokens=0`, the consistency branch rewrote `prompt_tokens = 0 - 5 = -5`, and the assertion caught `Usage(completion_tokens=5, prompt_tokens=-5, total_tokens=0)` on its way into `completion_cost`.

**Pass-after** — full PR branch: `8 passed in 25.41s`. Baseline: `main`'s own 7 tests in this file pass in this env, so the failure above is introduced by the new test, not by the environment.

**One nuance for the record**: the regression test can only exercise the `== 0` half of the guard, because litellm normalizes at construction:

```
>>> from litellm import Usage
>>> Usage(prompt_tokens=10, completion_tokens=5, total_tokens=None).total_tokens
0
```

The `TypeError` half of #935 is still reachable in production, though: `_query` returns the **portkey SDK's** response object, not litellm's, and that one preserves `None`:

```
>>> from portkey_ai.api_resources.types.chat_complete_type import Usage
>>> Usage(prompt_tokens=10, completion_tokens=5, total_tokens=None).total_tokens
None
```

The fix guards both (`total_tokens is None or total_tokens == 0`), so both mechanisms from the issue are covered — it's only the test that structurally can't reach the `None` branch while it builds a litellm `ModelResponse`. Pinning that branch would need a portkey-shaped usage stub; arguably not worth an extra test given the guard is a single condition.

Also checked: the diff leaves the cached-tokens rewrite branch (`total > prompt + completion`) untouched — it fires exactly as before; the new guard only runs ahead of it for the None/0 case.

arccat-114:
Hi, just a gentle follow-up when you have a chance — this PR is ready for review. No rush, and thanks for taking a look!

